Issue #187, opened 2020-05-09
Fix Timeline REST API: Concatenate older events to the feed when the current event log file is depleted.


ivelin:
Currently the event timeline API implementation only pulls events from the latest event log file. It should go to the next rollover event log as needed.

https://github.com/ambianic/ambianic-edge/blob/master/src/ambianic/webapp/server/samples.py#L152

ivelin:
Closed by #188 PR from  @muka

PR #188, opened 2020-05-10
feat: include rotated event logs in the timeline events

muka:
The implementation load the list of available logs and create a list of events order old-first. It may not perform well on a constrained device with a large number of files.

Refs #187

### DIFF
--- a/src/ambianic/webapp/server/samples.py
+++ b/src/ambianic/webapp/server/samples.py
@@ -144,15 +144,23 @@ def get_timeline(before_datetime=None, page=1, data_dir=None):
                         ' Error: %s', before_datetime, str(e))
     page_start_position = (page-1)*page_size
     page_end_position = page_start_position + page_size
+
     if not parsed_datetime:
         log.debug('Fetching most recent saved samples')
     log.debug('Fetching samples page %d. Page size %d. '
               'Sample index range [%d:%d]. ',
               page, page_size, page_start_position, page_end_position)
-    p = Path(data_dir) / 'timeline-event-log.yaml'
-    log.debug('Timeline path: %s', p.resolve())
-    with p.open() as pf:
-        timeline_events = yaml.safe_load(pf)
+
+    files = list(Path(data_dir).rglob('timeline-event-log.yaml*'))
+    files = sorted(files, key=os.path.getmtime, reverse=False)
+
+    # load the event history, older first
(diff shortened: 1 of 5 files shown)


src/ambianic/webapp/server/samples.py
```diff
@@ -144,36 +144,60 @@ def get_timeline(before_datetime=None, page=1, data_dir=None):
...
-    #    with open(json_file) as f:
-    #        sample = json.load(f)
-    #        sample['id'] = uuid.uuid4().hex
-    #        sample['file'] = str(json_file)
-    #        samples.append(sample)
-    # lines = map(str, files)
-    # log.debug('File names follow:\n %s', "\n".join(lines))
-    return timeline_slice
+
+    files = list(Path(data_dir).rglob('timeline-event-log.yaml*'))
```
ivelin, line 154: Very nice first cut @muka . Very much appreciate adding tests to keep our coverage high.

As you probably realized, event yaml files are just using the standard file rotation logger, which is supposed to ensure that the latest events are logged in the `timeline-event-log.yaml` file, the next most recent batch is in `timeline-event-log.yaml.1` and so on.

Based on that I think it would be sensible to load event files one at a time as needed using a file suffix increment to find and return the page requested by the API. Would you be interested in implementing such approach?

Loading all event files at ones makes the implementation even more inefficient than it already is. Over time the number of these files can grow substantially. 

I am still monitoring user feedback on performance, but at some point I suspect we will need to move to a time series database that can efficiently scan and return event ranges. I am open to fast forwarding this approach if you'd be more interested going that way.

muka: I did it, see the second commit. It should do what you described.

https://github.com/ambianic/ambianic-edge/commit/5a96034aed0ae0eb94e94f7ba68f49ecae6d827c

ivelin: You are fast!! Reviewing now...

ivelin:
closes #187

tests/timeline/timeline-event-log.yaml.0
```diff
@@ -0,0 +1,205 @@
+
```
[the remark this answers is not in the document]
muka: You are right, the default yaml is gitignored but I can --force add it

Issue #170, opened 2020-02-27
Rolling log files split between processes


ivelin:
It appears that when logs roll after reaching size limit, some processes write to the new log and others to the older log files.

muka:
Hi,
seems log rotation does not play nice with multithreading, see
https://docs.python.org/3/howto/logging-cookbook.html#logging-to-a-single-file-from-multiple-processes

`Although logging is thread-safe, and logging to a single file from multiple threads in a single process is supported, logging to a single file from multiple processes is not supported, because there is no standard way to serialize access to a single file across multiple processes in Python.`

A good replacement solution could be the [concurrent-log-handler](https://github.com/Preston-Landers/concurrent-log-handler) package.

ivelin:
That makes sense. The gstreamer pipeline elements run in separate processes.

PR #189, opened 2020-05-12
feat: use concurrent log rotation implementation

muka:
Uses [concurrent-log-handler](https://github.com/Preston-Landers/concurrent-log-handler) as drop-in replacement for native log rotation with better support for multithreading

Refs #170


ivelin:
Elegant solution! Thanks!🙏🏻

Closes #170

Issue #191, opened 2020-05-13
Reactive edge configuration


ivelin:
Implement live config.yaml file sync and reload on change.

Currently changes to config.yaml require restart. It is harder to implement robust config sync with proper graceful state transitions, but it may pay off long term as we move towards API and UI driven config changes.

See #190 as a related topic.

ivelin:
@muka does dynaconf resolve this feature request with https://github.com/ambianic/ambianic-edge/pull/222?

muka:
Not directly but there is a dedicated implementation with `wathcdog`. The implementation has been simplified so that the server will restart if the configuration file is changed on fs.

ivelin:
Then I will close this issue if you are OK with that @muka ?

muka:
Yes close thanks!

Issue #190, opened 2020-05-13
Expose REST API to add Camera Sources to Pipeline Config


ivelin:
Currently sources are configured in the config.yaml file which requires access to the Ambianic Edge file system and a restart.

As 98% of users are not technical enough nor interested in text configuration, we are gradually moving to a more UI driven approach.

A milestone towards that goal is to expose configuration options as REST API that can be then accessed from UI and other components (such as ONVIF scanner module).

ivelin:
I imagine the REST API would allow a remote client to add a new camera source with parameters:

id: unique identifier within an Ambianic installation (UUID4 may do)

display_name: user friendly display name

uri: where to pull data samples from 'http://...", "rtsp://..."

type: 'video', 'image', 'audio'
Hint that helps Ambianic Edge pick the most efficient way to pull data samples based on type (e.g. python requests for still images over http vs gstreamer for advanced rtsp streaming).

live: 'true', 'false'
whether to continue trying to reconnect and pull data even if there are disruptions and the data stream ends or errors.

ivelin:
As a side effect of introducing this API, I also imagine the config.yaml file would need to be updated with new values and file changes would have to be picked up by the rest of Ambianic Edge without restart.

ivelin:
Done.

PR #193, opened 2020-05-26
feat: configuration manager

muka:
- handles callback on change
- expose and api to work with centralized configurations
- watch for file changes

ref #191 and is a set of feature to be used in #190

### DIFF
--- a/src/ambianic/config_mgm.py
+++ b/src/ambianic/config_mgm.py
@@ -31,7 +31,8 @@ class ConfigurationManager:
     def stop(self):
         """Stop the config manager"""
         self.handlers = []
-        self.__config = None
+        with self.lock:
+            self.__config = None
         self.watch_stop()
         if self.watch_thread is not None:
             self.watch_thread.join()
@@ -49,9 +50,8 @@ class ConfigurationManager:
         """Watch for file changes"""
         inotify = INotify()
         wd = inotify.add_watch(self.work_dir, flags.MODIFY)
-
         while not self.watch_event.is_set():
-            for event in inotify.read(timeout=0):
+            for event in inotify.read(timeout=100, read_delay=100):
                 for filename in [self.CONFIG_FILE, self.SECRETS_FILE]:
                     if event.name == filename:
                         self.log.info("File change detected: %s", filename)
@@ -115,7 +115,9 @@ class ConfigurationManager:
 
             self.log.debug('loaded config from %r: %r',
                            self.CONFIG_FILE, config)
+
             return self.set(config)
+
         except FileNotFoundError:
             self.log.warning('Configuration file not found: %s', config_file)
             self.log.warning(
@@ -136,7 +138,8 @@ class ConfigurationManager:
             Returns a dictionary with current configurations.
 
         """
-        return self.__config
+        with self.lock:
+            return self.__config
 
     def set(self, new_config):
         """Set configuration
--- a/src/ambianic/config_mgm.py
+++ b/src/ambianic/config_mgm.py
@@ -2,7 +2,6 @@
 
 import os
 import logging
-from time import sleep
 import threading
 import yaml
 from inotify_simple import INotify, flags
--- a/src/ambianic/config_mgm.py
+++ b/src/ambianic/config_mgm.py
@@ -5,6 +5,7 @@ import logging
 from time import sleep
 import threading
 import yaml
+from inotify_simple import INotify, flags
 
 
 class ConfigurationManager:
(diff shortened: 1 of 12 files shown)


ivelin:
Very cool! I will start the review. 
Codecov shows above that coverage will drop a bit? Would you be able to address that?

muka:
Yes I will, tests rely on timers to detect changes but seems the implementation is buggy and need some more work

ivelin:
Maybe you could use mockups to control the timing elements.

ivelin:
@muka  How do you feel about using a yaml parser that tries to preserve comments? If it works well, it can be useful since the transition to all API & UI config may take some time. [Ruaml](https://pypi.org/project/ruamel.yaml/) is a parser I found but have not used. Claims to preserve comments on load and dump.
https://yaml.readthedocs.io/en/latest/overview.html

ivelin:
@muka I looked at the files changes. If I understand correctly, the current PR makes reactive the app logging and timeline event logging configuration. It leaves out the rest of the configuration elements such as sources, ai models and pipelines. Is that a correct interpretation?

src/ambianic/server.py
```diff
@@ -94,40 +93,30 @@ def _configure(env_work_dir=None):
...
+    config_manager.stop()
+
+    def logging_config_handler(config):
         # configure logging
         logging_config = None
         if config:
             logging_config = config.get('logging', None)
         _configure_logging(logging_config)
+
+    config_manager.register_handler(logging_config_handler)
```
ivelin, line 106: This registers regular app logging handler to react on changes to logging settings in config.yaml.

src/ambianic/server.py
```diff
@@ -94,40 +93,30 @@ def _configure(env_work_dir=None):
...
         timeline.configure_timeline(timeline_config)
-        return config
-    except FileNotFoundError:
-        log.warning('Configuration file not found: %s', config_file)
-        log.warning('Please provide a configuration file and restart.')
-    except Exception as e:
-        log.exception('Configuration Error!', e, exc_info=True)
-    return None
+
+    config_manager.register_handler(timeline_config_handler)
```
ivelin, line 115: this registers handler to update the timeline event logging settings if they change in config.yaml.

src/ambianic/server.py
ivelin: I do not see registering handler for the other sections in config yaml: sources, ai models, and pipelines. Is that left for a future PR? That is fine with me , we just need to document which parts of config.yaml are reactive and have hot reload capability and which ones require app restart.

src/ambianic/config_mgm.py
```diff
@@ -0,0 +1,163 @@
...
+        """Stop watching fs for changes"""
+        self.watch_event.set()
+
+    def save(self, config):
+        """Save configuration to file"""
+        if config is not None:
+            return
+
+        with open(self.get_config_file(), 'w') as fh:
+            yaml.dump(config, fh, default_flow_style=False)
```
ivelin, line 79: **Optional topic** related to the question I asked in a different comment. Can we try [Ruyaml](https://pypi.org/project/ruamel.yaml/) or another a drop-in replacement for PyYaml that preserves comments and structure of yaml file after load and save.
I think we will need a more granular API for config changes instead of a single YAML file dump over HTTP.
I am not sure if it will make sense to send back and forth the full yaml file between UI and edge. 
Full YAML file upload might make sense for fleet management for initial device setup, but after that, granular APIs would be probably more practical. I guess we will find out with time.

src/ambianic/server.py
muka: Yes, this PR aims at introducing the reactive configuration capabilities. I would go feature by feature to identify the right approach and expose a corresponding HTTP API.

I can do all in this PR but was thinking to have smaller PR with specific improvements.

Let me know your preferred approach

src/ambianic/server.py
muka: yes and get called every time. log handlers are resetted every time and thus seems to work fine. Eventually we'll need to understand if the handler instance must be explicitly disposed/closed to avoid possible leaks.

src/ambianic/config_mgm.py
muka: Totally agree, see also my comment above. I will go through the sections of the config and identify appropriate APIs to expose with validation.

src/ambianic/server.py
ivelin: Got it. I also like incremental PRs that are easy to comprehend and absorb. If anything regresses, which it never should if our coverage is good, we can quickly identify the problem, fix and add more test coverage. We can count this PR as phase 1 towards #191

PR #197, opened 2020-06-06
feat: reactive configuration

muka:
This PR introduce a "reactive configuration" by watching `dict` and `list` objects. Configuration may change via file or api (currently limited to `sources` section, but extensible). Changes are computed and applied in the pipeline server and in the main Ambianic server.

ref #191 #190


ivelin:
@muka Thank you. I am starting the review.

Do you agree with the CodeFactor suggestions:
```
 Check notice on line 30 in src/ambianic/pipeline/avsource/av_element.py

@codefactor-io
codefactor-io
/ CodeFactor

src/ambianic/pipeline/avsource/av_element.py#L30
Redefining built-in 'id' (redefined-builtin)
 Check notice on line 20 in src/ambianic/pipeline/ai/image_detection.py

@codefactor-io
codefactor-io
/ CodeFactor

src/ambianic/pipeline/ai/image_detection.py#L20
Redefining built-in 'id' (redefined-builtin)
 Check notice on line 30 in src/ambianic/pipeline/avsource/av_element.py

@codefactor-io
codefactor-io
/ CodeFactor

src/ambianic/pipeline/avsource/av_element.py#L30
Redefining built-in 'type' (redefined-builtin)
```

ivelin:
Looks like CodeCov is also complaining that the coverage went up to 93% earlier in the PR history then came down to 90%. But either way its up from the master, which is 86%. How would you like to address this one?

ivelin:
I see that you introduced an `id` sub-element for the elements of the sources section. Is that necessary? Since the sources section is a map with its keys required to be unique, I was thinking that the key is an ID itself and there is no need for a separate identifier. 

The intention is to allow user readable configuration files while making it possible to use via API as well.

I know you wanted to remove the use of yaml references, which is no big deal for me. But even in that case, elements can still refer to each other by their unique named keys in the config. Just don't prefix with yaml's own reference operators & or * , but do the semantic resolution in the config parser.

What am I missing?

muka:
Hi, to recap

- CodeFactor is complaining about `id` and I can rename those references to eg. `source_id`
- CodeCov is puzzling me as I added a bunch of new test but the coverage has not moved much forward. Seems `ai` and `avsource` package need more testing.
- The introduction of the `id` aim at supporting the lookup keeping backward-compatible to the references. In the current implementation, dict are not references thus each element become an autonomous object. Some logic may help in identify equality and have waterfall effects on changes. However it is not managed yet.

I would suggest to keep this PR open until we reach the desired functional capabilities. Thanks!

ivelin:
> Hi, to recap
> 
> * CodeFactor is complaining about `id` and I can rename those references to eg. `source_id`

OK, that would make sense contingent on the discussion below about the use of source id. 

> * CodeCov is puzzling me as I added a bunch of new test but the coverage has not moved much forward. Seems `ai` and `avsource` package need more testing.

Yes, I am ok with the 90%, which is obviously better than now. Its up to you if you'd like to add more tests beyond that.

> * The introduction of the `id` aim at supporting the lookup keeping backward-compatible to the references. In the current implementation, dict are not references thus each element become an autonomous object. Some logic may help in identify equality and have waterfall effects on changes. However it is not managed yet.

I am sorry, but I am not following your thoughst here. Could you please elaborate with some sample code. 

> 
> I would suggest to keep this PR open until we reach the desired functional capabilities. Thanks!

Sounds good. May you can switch it to a [draft PR](https://github.blog/2019-02-14-introducing-draft-pull-requests/) for now.

muka:
I seen in this project for the first time features of YAML such as `<<` and pointers and was nice to learn about.

Now that we are moving to an API for configuration, supporting such language specific feature can be challenging. 
Once the YAML is loaded and the API take care of updating some elements, how do we control pointers and merge keys while writing?

My proposal is to ensure 1:1 compatibility with a JSON file for example. I would use simple references to other sections eg. `section_id` or `model_id`.
Taking a pipeline as example, we monitor for `sources[source_id]` changes and trigger a reload in the pipeline. Similarly watching for a specific item by its id under `ai_models` ensure properties are picked up immediately by the pipeline.

Let me know if I can elaborate further

ivelin:
> I seen in this project for the first time features of YAML such as `<<` and pointers and was nice to learn about.
> 
> Now that we are moving to an API for configuration, supporting such language specific feature can be challenging.
> Once the YAML is loaded and the API take care of updating some elements, how do we control pointers and merge keys while writing?
> 
> My proposal is to ensure 1:1 compatibility with a JSON file for example. I would use simple references to other sections eg. `source` or `model`. See [an example](https://gist.github.com/muka/8221e733ab9ba9c37d8580ded5697f0c)
> Taking a pipeline as example, we monitor for `sources[source_id]` changes and trigger a reload in the pipeline. Similarly watching for a specific item by its id under `ai_models` ensure properties are picked up immediately by the pipeline.
> 
> Let me know if I can elaborate further

OK, it all makes perfect sense now. You are looking for a more flexible way to represent the configuration data in various formats. When we are ready to move to a database storage, we won't have to worry about mapping YAML specific constructs.

Go for it!

muka:
I would go on with review and merge if everything is ok. 

We will reach better coverage in another PR as I think some API under `webapp.server.samples` are still incomplete.

Running pytest with coverage we have a 93%  of overall covergage that is pretty respectable :)

src/ambianic/webapp/server/config_sources.py
```diff
@@ -0,0 +1,63 @@
...
+    config = config_manager.get()
+    if source_id not in config["sources"].keys():
+        raise NotFound("source id not found")
+
+    return config["sources"][source_id]
+
+
+def remove(source_id):
+    """Remove source by id"""
+    get(source_id)
```
ivelin, line 53: Do you mind adding INFO log statements for each configuration change operation. I'm sure they will come handy as we start testing API and UI config updates.

ivelin: Another thought that occured to me is the side effect of removing an id. Do we want to document what the expected behavior of the system is when a source is removed while the system is running?

src/ambianic/webapp/server/config_sources.py
```diff
@@ -0,0 +1,63 @@
...
+}
+
+source_types = ["video", "audio", "image"]
+
+
+def validate(source_id, source):
+    """Validate input object"""
+
+    if not isinstance(source, dict):
+        raise BadRequest("source should be an object")
```
ivelin, line 19: This is a bit of a generic error message. Maybe replace with something more specific like
`Source should be a valid dictionary of pipeline source objects'.

src/ambianic/webapp/server/config_sources.py
```diff
@@ -0,0 +1,63 @@
...
+    return config["sources"][source_id]
+
+
+def remove(source_id):
+    """Remove source by id"""
+    get(source_id)
+    del config_manager.get()["sources"][source_id]
+
+
+def save(source_id, source):
```
ivelin, line 57: This is just a stylistic comment, that I don't have strong feelings about. The word save usually implies persisting some data. In this case we are just adding an element to a dict. Should we use 'put' or 'add' as method name?

ivelin:
Do you have any further comments? 
I will proceed with the merge if I don't hear from you by tomorrow. I know my earlier comments were minor and optional so there is nothing blocking here.

src/ambianic/webapp/server/config_sources.py
muka: I use to use `save` when including `update` and `add`, however I am open to change it

src/ambianic/webapp/server/config_sources.py
muka: That's a good point, it will impact the pipeline services that will fail to reload in case of a missing id.

muka:
Hi, I am addressing your comments and adding further tests to handle the
case a source is removed and a pipeline has a reference to it. I will
update you once done, thanks!

muka:
I think we are good to go now

src/ambianic/config_mgm/__init__.py
```diff
@@ -0,0 +1,3 @@
+
+from .config_diff import ConfigChangedEvent, Config, ConfigList, ConfigDict, ConfigChangedEvent
```
ivelin, line 2: This is really minor, but it seems to make sense. ConfigChangeEvent is imported twice. I'll fix this.

PR #202, opened 2020-07-23
Update flaskr.py

dcmartin:
The relative path should be `.` not `./data`; the resolution being performed correctly is only when the name of the directory is also `/data`


ivelin:
@dcmartin this PR seems duplicate of #203 . Which one should we keep?

ivelin:
@dcmartin  Closing this draft PR due to inactivity. Let me know if you'd like to complete it.

PR #204, opened 2020-07-24
fix: sources API

muka:
This PR fixes the sources HTTP API reload capabilities. Functionally it is a step back as now all the pipelines will be restarted without fine grained checks.

Anyway we can recover such level of functionalities later on if required.


ivelin:
@muka Have you noticed that codecov is failing? Would you be able to add a few more tests to keep the coverage at a good level.

muka:
I noticed but considered just the overall, will check this

muka:
I think the PR is good for review. That codecov patch check is too picky for me!

ivelin:
Awesome. Merging. Thank you!

PR #205, opened 2020-07-31
feat: stream device support

muka:
Support for linux/osx local devices (eg. /dev/video0)

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -24,7 +24,7 @@ sources:
   #   type: video, audio, or auto
   #   type: video
   #   live: true
-  
+
   # local video device integration example
   webcam:
     uri: /dev/video0
--- a/config.yaml
+++ b/config.yaml
@@ -18,15 +18,24 @@ timeline:
 # Cameras and other input data sources
 # Using Home Assistant conventions to ease upcoming integration
 sources:
-#  front_door_camera:
-#    uri: *secret_uri_front_door_camera
-    # type: video, audio, or auto
-#    type: video
-#    live: true
-#  entry_area_camera:
-#    uri: *secret_uri_entry_area_camera
-#    type: video
-#    live: true
+
+  # front_door_camera:
+  #   uri: *secret_uri_front_door_camera
+  #   type: video, audio, or auto
+  #   type: video
+  #   live: true
+  
+  # local video device integration example
+  webcam:
+    uri: /dev/video0
+    type: video
+    live: true
+
+  um:uuid:5f5a69c2-e0ae-504f-829b-0076E58E0DDE:
+    live: true
+    type: video
+    uri: rtsp://192.168.7.13:554/11
+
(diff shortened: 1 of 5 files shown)


config.yaml
```diff
@@ -18,15 +18,24 @@ timeline:
...
+
+  # front_door_camera:
+  #   uri: *secret_uri_front_door_camera
+  #   type: video, audio, or auto
+  #   type: video
+  #   live: true
+
+  # local video device integration example
+  webcam:
+    uri: /dev/video0
```
ivelin, line 30: Which type of device and OS is this tested to work on? Any Linux, Mac, Win?
Can you please add a few lines to the [config docs](https://docs.ambianic.ai/users/configure/) to set user expectations.

config.yaml
```diff
@@ -18,15 +18,24 @@ timeline:
 # Cameras and other input data sources
 # Using Home Assistant conventions to ease upcoming integration
 sources:
-#  front_door_camera:
```
ivelin, line 21: OK, we can switch default to a local camera. Almost all dev computers nowadays have one. The question is whether all of them will work with /dev/video0 or we need a higher level abstraction like uri:default_local_camera and then parse that into something specific to the current environment. That would be my expectation if I was a new user. Usually apps don't ask the user how to local the default camera. They work with the hosting OS to do so.

tests/pipeline/avsource/test_gstreamer.py
```diff
@@ -480,3 +484,28 @@ def test_gst_set_state_playing_no_preroll():
...
     gst._gst_loop()
     assert gst.source.is_live
+
+
+def test_dev_video_source():
+    
+    def create_gst(fmt=None):
+        gst = GstService(
+            source_conf={
+                "uri": "/dev/video0",
```
ivelin, line 494: Does this test past in the cloud CI? Where would it find /dev/video0 ?

config.yaml
muka: You are right, focusing on linux (because of v4l2). It works with picamera also if the [kernel module is enabled](https://raspberrypi.stackexchange.com/a/26386/41277).

I do not have a windows / mac machine, I suspect won't work out of the box anyway

config.yaml
muka: Yup, this is a good idea indeed. Need to focus a bit more on OS detection in that case if gst does not have a "trick" for that. Anyway would be better to test on win and mac first, see also my comment above

tests/pipeline/avsource/test_gstreamer.py
muka: This test the parsing only, not the actually input. Setting up the transmission [seems feasible](https://stackoverflow.com/questions/18011304/how-to-make-a-virtual-video-device-for-testing-and-debugging-which-has-dev-vide) I think but could require some amount of work

PR #208, opened 2020-08-10
fix: failing fs watch test

ivelin:


ivelin:
The -u fix did not work. @muka is looking at replacing INotify with watchtower for file system monitoring.

PR #214, opened 2020-09-11
feat: initial picamera support

muka:
Initial support for picamera driver. Currently seems there is a misalignment with the GPIO python module for the PI4 chipset which causes an exception in docker that follows

```sh
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.7/dist-packages/picamera/__init__.py", line 89, in <module>
    from picamera.camera import PiCamera
  File "/usr/local/lib/python3.7/dist-packages/picamera/camera.py", line 79, in <module>
    from RPi import GPIO
  File "/usr/lib/python3/dist-packages/RPi/GPIO/__init__.py", line 23, in <module>
    from RPi._GPIO import *
RuntimeError: This module can only be run on a Raspberry Pi!
```

I am investigating how to manage the issue

### DIFF
--- a/ambianic-debug.sh
+++ b/ambianic-debug.sh
@@ -27,6 +27,11 @@ docker pull ambianic/ambianic-edge:dev
 docker run -it --rm \
   --name ambianic-edge-dev \
   --mount type=bind,source="$MY_DIR",target=/workspace \
+  -v /opt/vc:/opt/vc \
+  -v /sys/class/gpio:/sys/class/gpio \
+  -v /dev:/dev \
+  --device /dev/gpiomem \
+  --privileged \
   --net=host \
   $VIDEO_ARG \
   $USB_ARG \
--- a/ambianic-debug.sh
+++ b/ambianic-debug.sh
@@ -28,9 +28,6 @@ docker run -it --rm \
   --name ambianic-edge-dev \
   --mount type=bind,source="$MY_DIR",target=/workspace \
   -v /opt/vc:/opt/vc \
-  -v /sys/class/gpio:/sys/class/gpio \
-  -v /dev:/dev \
-  --device /dev/gpiomem \
   --privileged \
   --net=host \
   $VIDEO_ARG \
--- a/ambianic-debug.sh
+++ b/ambianic-debug.sh
@@ -15,6 +15,14 @@ else
   USB_ARG=""
 fi
 
+
+VC_SHARED_LIB=""
+if grep -s -q "Raspberry Pi" /proc/cpuinfo;
+then
+  VC_SHARED_LIB="-v /opt/vc:/opt/vc"
+fi
+
+
 if [ -e "$VIDEO_0" ]; then
   VIDEO_ARG="--device $VIDEO_0"
 else
@@ -27,7 +35,7 @@ docker pull ambianic/ambianic-edge:dev
 docker run -it --rm \
   --name ambianic-edge-dev \
   --mount type=bind,source="$MY_DIR",target=/workspace \
-  -v /opt/vc:/opt/vc \
+  $VC_SHARED_LIB \
   --privileged \
   --net=host \
   $VIDEO_ARG \
--- a/build/install_requirements.sh
+++ b/build/install_requirements.sh
@@ -50,16 +50,22 @@ if $(arch | grep -q arm)
 # there is no RPI firmware in docker images, so we will install on ARM flag
 #if grep -s -q "Raspberry Pi" /sys/firmware/devicetree/base/model; then
 then
-  echo "Installing Raspberry Pi / ARM CPU specific dependencies"
-  sudo pip3 install --upgrade RPi.GPIO picamera
+  # avoid install during CI builds
+  if [ ! $TRAVIS ] ; then
+    echo "Installing Raspberry Pi / ARM CPU specific dependencies"
+    sudo pip3 install --upgrade RPi.GPIO picamera
+  fi
+
   # sudo apt-get install -y modprobe
   # Add v4l2 video module to kernel
   #  if ! grep -q "bcm2835-v4l2" /etc/modules; then
   #    echo bcm2835-v4l2 | sudo tee -a /etc/modules
   #  fi
   #  sudo modprobe bcm2835-v4l2
+
   # Enable python wheels for rpi
   sudo cp raspberrypi.pip.conf /etc/pip.conf
+  
 fi
 
 # install python dependencies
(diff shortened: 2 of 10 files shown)


ivelin:
@muka this PR is marked as ready? Did you mean to mark it as draft. CI seems to be failing.

muka:
Actually seems travis timed out, will restart the build. Coverage has not changed as picamera requires hardware to start up.

ambianic-debug.sh
```diff
@@ -27,6 +27,8 @@ docker pull ambianic/ambianic-edge:dev
 docker run -it --rm \
   --name ambianic-edge-dev \
   --mount type=bind,source="$MY_DIR",target=/workspace \
+  -v /opt/vc:/opt/vc \
```
ivelin, line 30: Would /opt/vc work on a non-raspberry pi machine? Do we need to add conditional logic in the shell file for this parameter?

src/ambianic/pipeline/avsource/av_element.py
```diff
@@ -10,7 +10,8 @@
 import requests
 from ambianic.util import stacktrace
 from ambianic.pipeline import PipeElement
-from ambianic.pipeline.avsource import gst_process
+from ambianic.pipeline.avsource import gst_process, picam
+from ambianic.pipeline.avsource.picam import Picamera
```
ivelin, line 14: Would this Picamera import work on a non-raspberry pi machine? Do we need to include conditional logic to only enable and make use of picam libs if its running on rpi?

src/ambianic/pipeline/avsource/av_element.py
```diff
@@ -104,6 +105,22 @@ def _fetch_img_exception_recovery(self):
...
+    def _run_picamera_fetch(self):
+
+        picamera = Picamera()
+        if picamera.has_failure():
+            log.error("Picamera setup failed, exiting..")
+            return None
+
+        while not self._stop_requested:
+            try:
+                self.receive_next_sample(image=picamera.acquire())
```
ivelin, line 117: Does it make sense to add a few coverage tests for this section of code that codecov is complaining about?  The _TestPiCamera  mockup of the Picamera class that you already use for other unit tests should be still applicable here.

src/ambianic/pipeline/avsource/picam.py
```diff
@@ -0,0 +1,54 @@
...
+            except Exception as err:
+                log.warn("Error importing picamera module: %s" % err)
+                self.error = err
+                return
+        else:
+            self.camera = picamera_override.PiCamera()
+
+        self.stream = BytesIO()
+        self.camera.led = True
+        self.camera.start_preview()
```
ivelin, line 35: I don't think start_preview() is applicable for us. It asks the picam library to display camera feed to the local device display, which in our case doesn't usually exist.

src/ambianic/pipeline/avsource/av_element.py
```diff
@@ -104,6 +105,22 @@ def _fetch_img_exception_recovery(self):
...
+            except Exception as err:
+                log.debug("Error acquiring from picamera: %s" % err)
+                time.sleep(1)
+
+        picamera.stop()
```
ivelin, line 122: Similar comment as above related to the codecov flag.

ivelin (commented):
Awesome stuff @muka . Added a few comments below for your review.

src/ambianic/pipeline/avsource/av_element.py
muka: `picam ` is a wrapper for the "real" picamera and will return an error flag if picamera import fails

ambianic-debug.sh
muka: yes you are right, will provide it

src/ambianic/pipeline/avsource/av_element.py
muka: There is a test for this in test_av_element but seems it is not working, will check again

src/ambianic/pipeline/avsource/picam.py
muka: good point, thanks

muka:
Test have been addressed properly, there is still an issue in the build pipe for armv6 that seems non able to complete. Moving forward the PR and eventually open a dedicated issue to support older PI?

build/install_requirements.sh
```diff
@@ -63,9 +67,12 @@ then
   sudo cp raspberrypi.pip.conf /etc/pip.conf
 fi
 
+
+# install cython v0.29 to address a compile time bug for h5py
+# see https://github.com/h5py/h5py/issues/1533#issuecomment-617519250
+sudo pip3 install cython
```
ivelin, line 73: Do we really need cython for picamera? I've used the picam lib for a number of projects without cython.

ivelin: h5py is also available as a binary from the rpi binary repo. We've been downloading it without issue. No need to build from source.

build/install_requirements.sh
```diff
@@ -39,20 +39,24 @@ sudo apt-get install -y libgstreamer1.0-0 gstreamer1.0-plugins-base \
...
 
+pip3 --version
+sudo pip3 install -U pip
+
 # Install Raspberry Pi video drivers
-if $(arch | grep -q arm)
+# if $(arch | grep -q arm)
 # there is no RPI firmware in docker images, so we will install on ARM flag
 #if grep -s -q "Raspberry Pi" /sys/firmware/devicetree/base/model; then
+if grep -s -q "Raspberry Pi" /proc/cpuinfo;
```
ivelin, line 56: It looks like this check is failing. I noticed you commented out the earlier arm check and replaced by a Raspibery Pi check. However in the CI environment this fails and the whole section with rpi specific deps does not run.

build/install_requirements.sh
```diff
@@ -39,20 +39,24 @@ sudo apt-get install -y libgstreamer1.0-0 gstreamer1.0-plugins-base \
...
 # make sure python sees the packages installed via apt-get
 # export PYTHONPATH=$PYTHONPATH:/usr/lib/python3/dist-packages
 # echo "export PYTHONPATH=$PYTHONPATH:/usr/lib/python3/dist-packages" >> $HOME/.bashrc
 
+pip3 --version
+sudo pip3 install -U pip
+
 # Install Raspberry Pi video drivers
-if $(arch | grep -q arm)
+# if $(arch | grep -q arm)
```
ivelin, line 53: Commenting out this arm check fails the CI run for arm.

ivelin (commented):
See new comments related to arm CI failure.

build/install_requirements.sh
muka: No indeed, it has fallen in as a cascade of bug I created!

build/install_requirements.sh
muka: I think this is the line that started up the chain of build error. For some reason I changed it to match just RPi. Good catch!

ivelin:
Awesome @muka ! Thank you!!
Could you please push an update to docs with some minimal instructions how this works?

PR #217, opened 2020-09-27
Fix: use picamera continous_capture

muka:


ivelin (approved):
You discovered an interesting problem that I haven't ran into before, but I see there are forum discussion. When you recycle the PiCamera object too fast there seem to be racing conditions with the GPIO cleanup code on camera.close(). Maintaining a long running picamera object with a separate queue of images to pull from seems to be a sensible solution.

PR #222, opened 2020-11-14
feat: config api refactor

muka:


ivelin:
Good stuff @muka . Got it and will do a review.

I don't understand why codecov shows increase in coverage up in the PR log, but at the check level below it shows decrease.

ivelin (approved):
Looking good. Merging.

PR #225, opened 2020-11-24
fix: remove ca install on start

ivelin:
test if still needed with latest rpi os


ivelin (commented):
test if new rpi os fixed ca issues in docker

PR #232, opened 2020-12-05
feat: fall detection using posenet model

bhavikapanara:


ivelin (commented):
Hi @bhavikapanara . Great milestone. I will approve this PR as a good first step in the fall detection process. I will also post a few comments inline for your consideration and discussion to be sure we are on the same page.

src/ambianic/pipeline/ai/fall_detect.py
```diff
@@ -0,0 +1,110 @@
...
+            if not self._prev_vals:
+                res = False
+            else:
+                temp_left_point = [[self._prev_vals[0][0], self._prev_vals[0][1]], [pose_vals_list[0][0], pose_vals_list[0][1]]]
+                left_angle = self.calculate_angle(temp_left_point)
+
+                temp_right_point = [[self._prev_vals[1][0], self._prev_vals[1][1]], [pose_vals_list[1][0], pose_vals_list[1][1]]]
+                rigth_angle = self.calculate_angle(temp_right_point)
+
+                if left_angle > self._fall_factor or rigth_angle > self._fall_factor:
```
ivelin, line 105: @bhavikapanara since fall factor is preset to 20 degrees, the test here checks whether the body line is tilted 20' off vertical position. Is that correct? 

I see in the test that you've chosen a person sitting straight up and another person who is crouching and maybe about to fall, but not on the floor. From the image it is not clear if the person will fall or squat back up.

Should we pick a bigger angle, maybe 60 to have a reasonably high level of confidence that its a fall.

I understand that if the angle is bigger, the pose model may not detect correctly the body points, which is why we discussed the approach of rotating the image +/- 90' in such cases to see if you can get better pose estimates.

In a follow up project (not this PR), we will discuss re-training the model with more data using rotation during training so that the model is more aware of fallen positions and won't require rotation in the inference stage later on.

Looking forward to your thoughts.

src/ambianic/pipeline/ai/fall_detect.py
```diff
@@ -0,0 +1,110 @@
...
+    def calculate_angle(self, p):
+        x1, y1 = p[0][0][0], p[0][0][1]
+        x2, y2 = p[0][1][0], p[0][1][1]
+        x3, y3 = p[1][0][0], p[1][0][1]
+        x4, y4 = p[1][1][0], p[1][1][1]
+        
+        m1 = (y1-y2)/(x1-x2)
+        m2 = (y3-y4)/(x3-x4)
+        
+        theta1 = math.degrees(math.atan(m1))
```
ivelin, line 70: @bhavikapanara do you take into account situations when the angle may be a negative number but still meets our criteria? 

Also, thinking some more about it, do we want the angle in the first image to be close to 90' (+/- 15') (implying the person is standing up) and then the angle in the second image to be more than +/- 60' relative to the first one? The idea here is to prevent situations where a person is lying down and they stand up, which will be a more than 60' change of angle but they would be just getting off a couch or a bed, instead of falling down or laying down.

I don't think we need to go too far with these empirical rules, because it will turn into a spaghetti code. Instead we can start a new project (after this PR) to design an RNN that predicts fall based on a sequence of poses.

Your thoughts will be much appreciated.

src/ambianic/pipeline/ai/fall_detect.py
```diff
@@ -72,39 +72,63 @@ def calculate_angle(self, p):
...
         return theta
+    
+    def find_keypoints(self, image):
+
+        image_ori = image.copy()
+        rotation = [90, -90]
+        
+        while True:
+            poses = self._pose_engine.DetectPosesInImage(image)
+            for pose in poses:
```
ivelin, line 83: @bhavikapanara is there a chance that poses is None  or empty? In such cases this while True loop will spin forever, no?

tests/pipeline/ai/test_fall_detect.py
```diff
@@ -79,6 +79,28 @@ def sample_callback(image=None, inference_result=None, **kwargs):
...
+        result = inference_result
+
+    fall_detector = FallDetector(**config)
+
+    output = _OutPipeElement(sample_callback=sample_callback)
+    
+    fall_detector.connect_to_next_element(output)
+
     img_1 = _get_image(file_name='fall_img_1.png')
     img_2 = _get_image(file_name='fall_img_2.png')
```
ivelin, line 105: @bhavikapanara I see that fall_img_2.png is a person leaning over. Is the angle between standing and leaning over 60'? Could you please attache images with pose lines overlay in each image?

Let's add a negative test case that covers images where the poses are detected but the angle is less than the required fall factor.

ivelin (changes requested):
@bhavikapanara you are on fire! See a few minor comments. 

We should be able to merge by tomorrow.

src/ambianic/pipeline/ai/fall_detect.py
bhavikapanara: No...there are no chances to get an empty result.
that will always return key-points. If there is no person in a frame, it will return key-points with very very low probability score.

I have checked for this image [no person](https://github.com/ambianic/ambianic-edge/blob/fall-detect/tests/pipeline/ai/background.jpg) image. below is the return response from posenet.


`[Pose({'nose': Keypoint(<nose>, [1335, 171], 0.0018767025657813801), 'left eye': Keypoint(<left eye>, [1295, 6], 0.0011504802784690327), 'right eye': Keypoint(<right eye>, [1305, 11], 0.0015948671443740245), 'left ear': Keypoint(<left ear>, [324, 445], 0.003422743024364482), 'right ear': Keypoint(<right ear>, [478, 347], 0.0011285191402948904), 'left shoulder': Keypoint(<left shoulder>, [1160, 499], 0.006819241507138113), 'right shoulder': Keypoint(<right shoulder>, [1116, 487], 0.0022483134904298474), 'left elbow': Keypoint(<left elbow>, [672, 754], 0.011247683477665103), 'right elbow': Keypoint(<right elbow>, [1071, 552], 0.0051987406127019156), 'left wrist': Keypoint(<left wrist>, [588, 734], 0.013236690572867188), 'right wrist': Keypoint(<right wrist>, [403, 653], 0.008069606264386485), 'left hip': Keypoint(<left hip>, [1111, 740], 0.0112793711392602), 'right hip': Keypoint(<right hip>, [1066, 737], 0.01047226251029573), 'left knee': Keypoint(<left knee>, [638, 496], 0.0036192633880051815), 'right knee': Keypoint(<right knee>, [468, 521], 0.0037731794967397485), 'left ankle': Keypoint(<left ankle>, [488, 630], 0.03046229198464312), 'right ankle': Keypoint(<right ankle>, [473, 628], 0.03632458557985564)}, 0.0)]
`

Here, in the response, last value **`0.0`** is the pose score, it depends on a number of key-points are detected with more than 60% probability score.

ivelin: That is good. As a best practice, it's good to account for edge cases that may put your code in an infinite loop. I would still add a check to exit the loop in case there are no poses returned. You never know if the model has a bug or the next version of the model trims results below a given confidence threshold of top-k threshold. Either way infinite blocking  is not desirable. Better to have some sort of exit - either positive or an error report.

bhavikapanara: Yes...sure. I think it is best practise to consider all scenario. Sorry, I forgot it.
Will update the code

tests/pipeline/ai/test_fall_detect.py
ivelin: @bhavikapanara thoughts on this one?

src/ambianic/pipeline/ai/fall_detect.py
ivelin: Great. Thank you!

tests/pipeline/ai/test_fall_detect.py
bhavikapanara: Yes...just commit negative test case

src/ambianic/pipeline/ai/fall_detect.py
ivelin: Thank you for the update @bhavikapanara . Closing this discussion.

src/ambianic/pipeline/ai/fall_detect.py
ivelin: Thanks for the fix. Closing.

tests/pipeline/ai/test_fall_detect.py
bhavikapanara: fall_img_1
fall_img_2

Here are the key-points of the shoulder and hip : [[left shoulder, left hip], [right shoulder, right hip]]

First Image      :  `[[[436, 148], [431, 211]], [[401, 151], [416, 217]]]`
Second Image :  `[[[309, 209], [336, 278]], [[274, 211], [306, 280]]]`

ivelin: OK, this looks right. Seems like the angle is less than 60', no? What number do you get with your formula?

bhavikapanara: I also have doubts. Can you please look at this [Colab file](https://colab.research.google.com/drive/1UdrVwxJb3v6dqmPNmD5_UxrJQPo0VzGF?usp=sharing)?

ivelin: OK, let me look closer. Something is off with the calculations.

ivelin: OK, please look at the suggested corrections in the colab file and see you get better results on your end. The main issue was with using atan which does not give us the correct quadrant of an angle, whereas atan2 does.

More here: https://stackoverflow.com/questions/2676719/calculating-the-angle-between-the-line-defined-by-two-points

ivelin: The main change:

```
import math

def calculate_angle(p):
    x1, y1 = p[0][0][0], p[0][0][1]
    x2, y2 = p[0][1][0], p[0][1][1]
    x3, y3 = p[1][0][0], p[1][0][1]
    x4, y4 = p[1][1][0], p[1][1][1]
    
    print("x1, y1 : ",x1, y1)
    print("x2, y2 : ", x2, y2)
    print("x3, y3 : ", x3, y3)
    print("x4, y4 : ", x4, y4)
    
    # since y coordinates in screen and pose estimation are reversed
    # (higher values are lower on the screen), we need to reverse the y delta too

    theta1 = math.degrees(math.atan2(y2-y1, x1-x2))
    theta2 = math.degrees(math.atan2(y4-y3, x3-x4))
        
    print("theta1 : ", theta1)
    print("theta2 : ", theta2)

    theta = abs(theta1-theta2)
    return theta
```

Issue #237, opened 2020-12-11
[BUG] fall detection angle calculation


ivelin:
The calculation between two posenet body lines has a bug.

See context here with incorrectly passing test and a solution:
https://github.com/ambianic/ambianic-edge/pull/232#discussion_r540481310

PR #238, opened 2020-12-11
feat: fall detection-Improved angle calculation

bhavikapanara:


src/ambianic/pipeline/ai/fall_detect.py
```diff
@@ -63,12 +63,9 @@ def calculate_angle(self, p):
...
         x3, y3 = p[1][0][0], p[1][0][1]
         x4, y4 = p[1][1][0], p[1][1][1]
-        
-        m1 = (y1-y2)/(x1-x2)
-        m2 = (y3-y4)/(x3-x4)
-        
-        theta1 = math.degrees(math.atan(m1))
-        theta2 = math.degrees(math.atan(m2))
+
+        theta1 = math.degrees(math.atan2(y2-y1, x1-x2))
```
ivelin, line 67: Thank you @bhavikapanara . This update closes https://github.com/ambianic/ambianic-edge/issues/237
I am migrating the CI from Travis to Github Action today so we can run a full CI cycle  and push a new release.

bhavikapanara: ok.
Thanks, @ivelin

ivelin:
@bhavikapanara Noticed that you replaced image_2 in one of the tests to ensure that it passes'. Can you please put back the earlier image with the person leaning slightly and use that in a negative test. The goal is to check that when we have a difference of angles less than fall_factor, there is no fall_detection triggered. In addition the test will verify that our angle calculation is implemented correctly. Thank you!

PR #270, opened 2020-12-16
fix: fall detection bugs

bhavikapanara:


src/ambianic/pipeline/ai/fall_detect.py
```diff
@@ -35,6 +35,7 @@ def __init__(self,
 
         self._pose_engine = PoseEngine(self._tfengine)
         self._fall_factor = 60
+        self.pose_confidence_threshold = kwargs['confidence_threshold']
```
ivelin, line 38: What if confidence_threshold is not provided as a config argument? 

As a best practice, the code should assume reasonable default values.

Since FallDetector extends TFImageDetector, the confidence_threshold variable is already instantiated after the `super().__init__(model, **kwargs)` call. So there is no need to declare another instance variable with the same use.

However we can move some of the code from ImageDetection that assumes that the model returns a box down to ObjectDetection. That would keep ImageDetector cleaner and more relevant to FallDetector. 

With that, we can also move the default values for confidence away from ImageDetector, since its too abstract and early to assume 0.6 is a good default. Instead the default value could go down to ObjectDetector and FallDetector respectively where we have a better idea what confidence defaults work well.

ivelin (changes requested):
Nice work! 

I will now take the opportunity to gradually introduce you to the rest of the ML backend. See comments below.

ivelin:
> Hello @bhavikapanara! Thanks for updating this PR. We checked the lines you've touched for [PEP 8](https://www.python.org/dev/peps/pep-0008) issues, and found:
> 
> * In the file [`src/ambianic/pipeline/ai/fall_detect.py`](https://github.com/ambianic/ambianic-edge/blob/8c6d0d2ca22e5c72f30dd5e58324d162f085de33/src/ambianic/pipeline/ai/fall_detect.py):
> 
> > [Line 35:80](https://github.com/ambianic/ambianic-edge/blob/8c6d0d2ca22e5c72f30dd5e58324d162f085de33/src/ambianic/pipeline/ai/fall_detect.py#L35): [E501](https://duckduckgo.com/?q=pep8%20E501) line too long (106 > 79 characters)
> > [Line 36:80](https://github.com/ambianic/ambianic-edge/blob/8c6d0d2ca22e5c72f30dd5e58324d162f085de33/src/ambianic/pipeline/ai/fall_detect.py#L36): [E501](https://duckduckgo.com/?q=pep8%20E501) line too long (107 > 79 characters)
> > [Line 94:1](https://github.com/ambianic/ambianic-edge/blob/8c6d0d2ca22e5c72f30dd5e58324d162f085de33/src/ambianic/pipeline/ai/fall_detect.py#L94): [W293](https://duckduckgo.com/?q=pep8%20W293) blank line contains whitespace
> > [Line 97:1](https://github.com/ambianic/ambianic-edge/blob/8c6d0d2ca22e5c72f30dd5e58324d162f085de33/src/ambianic/pipeline/ai/fall_detect.py#L97): [W293](https://duckduckgo.com/?q=pep8%20W293) blank line contains whitespace
> > [Line 100:1](https://github.com/ambianic/ambianic-edge/blob/8c6d0d2ca22e5c72f30dd5e58324d162f085de33/src/ambianic/pipeline/ai/fall_detect.py#L100): [W293](https://duckduckgo.com/?q=pep8%20W293) blank line contains whitespace
> > [Line 103:1](https://github.com/ambianic/ambianic-edge/blob/8c6d0d2ca22e5c72f30dd5e58324d162f085de33/src/ambianic/pipeline/ai/fall_detect.py#L103): [W293](https://duckduckgo.com/?q=pep8%20W293) blank line contains whitespace
> > [Line 174:80](https://github.com/ambianic/ambianic-edge/blob/8c6d0d2ca22e5c72f30dd5e58324d162f085de33/src/ambianic/pipeline/ai/fall_detect.py#L174): [E501](https://duckduckgo.com/?q=pep8%20E501) line too long (90 > 79 characters)
> > [Line 175:80](https://github.com/ambianic/ambianic-edge/blob/8c6d0d2ca22e5c72f30dd5e58324d162f085de33/src/ambianic/pipeline/ai/fall_detect.py#L175): [E501](https://duckduckgo.com/?q=pep8%20E501) line too long (91 > 79 characters)
> > [Line 188:80](https://github.com/ambianic/ambianic-edge/blob/8c6d0d2ca22e5c72f30dd5e58324d162f085de33/src/ambianic/pipeline/ai/fall_detect.py#L188): [E501](https://duckduckgo.com/?q=pep8%20E501) line too long (274 > 79 characters)
> > [Line 191:80](https://github.com/ambianic/ambianic-edge/blob/8c6d0d2ca22e5c72f30dd5e58324d162f085de33/src/ambianic/pipeline/ai/fall_detect.py#L191): [E501](https://duckduckgo.com/?q=pep8%20E501) line too long (88 > 79 characters)
> > [Line 193:80](https://github.com/ambianic/ambianic-edge/blob/8c6d0d2ca22e5c72f30dd5e58324d162f085de33/src/ambianic/pipeline/ai/fall_detect.py#L193): [E501](https://duckduckgo.com/?q=pep8%20E501) line too long (144 > 79 characters)
> > [Line 194:80](https://github.com/ambianic/ambianic-edge/blob/8c6d0d2ca22e5c72f30dd5e58324d162f085de33/src/ambianic/pipeline/ai/fall_detect.py#L194): [E501](https://duckduckgo.com/?q=pep8%20E501) line too long (82 > 79 characters)
> > [Line 195:80](https://github.com/ambianic/ambianic-edge/blob/8c6d0d2ca22e5c72f30dd5e58324d162f085de33/src/ambianic/pipeline/ai/fall_detect.py#L195): [E501](https://duckduckgo.com/?q=pep8%20E501) line too long (83 > 79 characters)
> > [Line 197:80](https://github.com/ambianic/ambianic-edge/blob/8c6d0d2ca22e5c72f30dd5e58324d162f085de33/src/ambianic/pipeline/ai/fall_detect.py#L197): [E501](https://duckduckgo.com/?q=pep8%20E501) line too long (85 > 79 characters)
> > [Line 200:80](https://github.com/ambianic/ambianic-edge/blob/8c6d0d2ca22e5c72f30dd5e58324d162f085de33/src/ambianic/pipeline/ai/fall_detect.py#L200): [E501](https://duckduckgo.com/?q=pep8%20E501) line too long (89 > 79 characters)
> > [Line 202:80](https://github.com/ambianic/ambianic-edge/blob/8c6d0d2ca22e5c72f30dd5e58324d162f085de33/src/ambianic/pipeline/ai/fall_detect.py#L202): [E501](https://duckduckgo.com/?q=pep8%20E501) line too long (145 > 79 characters)
> > [Line 203:80](https://github.com/ambianic/ambianic-edge/blob/8c6d0d2ca22e5c72f30dd5e58324d162f085de33/src/ambianic/pipeline/ai/fall_detect.py#L203): [E501](https://duckduckgo.com/?q=pep8%20E501) line too long (84 > 79 characters)
> > [Line 204:80](https://github.com/ambianic/ambianic-edge/blob/8c6d0d2ca22e5c72f30dd5e58324d162f085de33/src/ambianic/pipeline/ai/fall_detect.py#L204): [E501](https://duckduckgo.com/?q=pep8%20E501) line too long (84 > 79 characters)
> > [Line 206:80](https://github.com/ambianic/ambianic-edge/blob/8c6d0d2ca22e5c72f30dd5e58324d162f085de33/src/ambianic/pipeline/ai/fall_detect.py#L206): [E501](https://duckduckgo.com/?q=pep8%20E501) line too long (85 > 79 characters)
> > [Line 210:80](https://github.com/ambianic/ambianic-edge/blob/8c6d0d2ca22e5c72f30dd5e58324d162f085de33/src/ambianic/pipeline/ai/fall_detect.py#L210): [E501](https://duckduckgo.com/?q=pep8%20E501) line too long (86 > 79 characters)
> > [Line 211:80](https://github.com/ambianic/ambianic-edge/blob/8c6d0d2ca22e5c72f30dd5e58324d162f085de33/src/ambianic/pipeline/ai/fall_detect.py#L211): [E501](https://duckduckgo.com/?q=pep8%20E501) line too long (88 > 79 characters)
> > [Line 213:80](https://github.com/ambianic/ambianic-edge/blob/8c6d0d2ca22e5c72f30dd5e58324d162f085de33/src/ambianic/pipeline/ai/fall_detect.py#L213): [E501](https://duckduckgo.com/?q=pep8%20E501) line too long (93 > 79 characters)
> > [Line 213:91](https://github.com/ambianic/ambianic-edge/blob/8c6d0d2ca22e5c72f30dd5e58324d162f085de33/src/ambianic/pipeline/ai/fall_detect.py#L213): [E202](https://duckduckgo.com/?q=pep8%20E202) whitespace before ')'
> > [Line 217:80](https://github.com/ambianic/ambianic-edge/blob/8c6d0d2ca22e5c72f30dd5e58324d162f085de33/src/ambianic/pipeline/ai/fall_detect.py#L217): [E501](https://duckduckgo.com/?q=pep8%20E501) line too long (137 > 79 characters)
> > [Line 221:49](https://github.com/ambianic/ambianic-edge/blob/8c6d0d2ca22e5c72f30dd5e58324d162f085de33/src/ambianic/pipeline/ai/fall_detect.py#L221): [E221](https://duckduckgo.com/?q=pep8%20E221) multiple spaces before operator
> 
> * In the file [`src/ambianic/pipeline/ai/pose_engine.py`](https://github.com/ambianic/ambianic-edge/blob/8c6d0d2ca22e5c72f30dd5e58324d162f085de33/src/ambianic/pipeline/ai/pose_engine.py):
> 
> > [Line 63:80](https://github.com/ambianic/ambianic-edge/blob/8c6d0d2ca22e5c72f30dd5e58324d162f085de33/src/ambianic/pipeline/ai/pose_engine.py#L63): [E501](https://duckduckgo.com/?q=pep8%20E501) line too long (92 > 79 characters)
> > [Line 89:80](https://github.com/ambianic/ambianic-edge/blob/8c6d0d2ca22e5c72f30dd5e58324d162f085de33/src/ambianic/pipeline/ai/pose_engine.py#L89): [E501](https://duckduckgo.com/?q=pep8%20E501) line too long (89 > 79 characters)
> > [Line 95:80](https://github.com/ambianic/ambianic-edge/blob/8c6d0d2ca22e5c72f30dd5e58324d162f085de33/src/ambianic/pipeline/ai/pose_engine.py#L95): [E501](https://duckduckgo.com/?q=pep8%20E501) line too long (108 > 79 characters)
> > [Line 124:80](https://github.com/ambianic/ambianic-edge/blob/8c6d0d2ca22e5c72f30dd5e58324d162f085de33/src/ambianic/pipeline/ai/pose_engine.py#L124): [E501](https://duckduckgo.com/?q=pep8%20E501) line too long (82 > 79 characters)
> > [Line 127:80](https://github.com/ambianic/ambianic-edge/blob/8c6d0d2ca22e5c72f30dd5e58324d162f085de33/src/ambianic/pipeline/ai/pose_engine.py#L127): [E501](https://duckduckgo.com/?q=pep8%20E501) line too long (90 > 79 characters)
> > [Line 130:80](https://github.com/ambianic/ambianic-edge/blob/8c6d0d2ca22e5c72f30dd5e58324d162f085de33/src/ambianic/pipeline/ai/pose_engine.py#L130): [E501](https://duckduckgo.com/?q=pep8%20E501) line too long (98 > 79 characters)
> > [Line 147:50](https://github.com/ambianic/ambianic-edge/blob/8c6d0d2ca22e5c72f30dd5e58324d162f085de33/src/ambianic/pipeline/ai/pose_engine.py#L147): [E231](https://duckduckgo.com/?q=pep8%20E231) missing whitespace after ','
> > [Line 147:67](https://github.com/ambianic/ambianic-edge/blob/8c6d0d2ca22e5c72f30dd5e58324d162f085de33/src/ambianic/pipeline/ai/pose_engine.py#L147): [E231](https://duckduckgo.com/?q=pep8%20E231) missing whitespace after ','
> 
> * In the file [`tests/pipeline/ai/test_fall_detect.py`](https://github.com/ambianic/ambianic-edge/blob/8c6d0d2ca22e5c72f30dd5e58324d162f085de33/tests/pipeline/ai/test_fall_detect.py):
> 
> > [Line 223:1](https://github.com/ambianic/ambianic-edge/blob/8c6d0d2ca22e5c72f30dd5e58324d162f085de33/tests/pipeline/ai/test_fall_detect.py#L223): [E302](https://duckduckgo.com/?q=pep8%20E302) expected 2 blank lines, found 1
> > [Line 224:80](https://github.com/ambianic/ambianic-edge/blob/8c6d0d2ca22e5c72f30dd5e58324d162f085de33/tests/pipeline/ai/test_fall_detect.py#L224): [E501](https://duckduckgo.com/?q=pep8%20E501) line too long (118 > 79 characters)
> > [Line 235:1](https://github.com/ambianic/ambianic-edge/blob/8c6d0d2ca22e5c72f30dd5e58324d162f085de33/tests/pipeline/ai/test_fall_detect.py#L235): [W293](https://duckduckgo.com/?q=pep8%20W293) blank line contains whitespace
> > [Line 242:16](https://github.com/ambianic/ambianic-edge/blob/8c6d0d2ca22e5c72f30dd5e58324d162f085de33/tests/pipeline/ai/test_fall_detect.py#L242): [E201](https://duckduckgo.com/?q=pep8%20E201) whitespace after '('
> > [Line 242:54](https://github.com/ambianic/ambianic-edge/blob/8c6d0d2ca22e5c72f30dd5e58324d162f085de33/tests/pipeline/ai/test_fall_detect.py#L242): [E202](https://duckduckgo.com/?q=pep8%20E202) whitespace before ')'
> > [Line 246:1](https://github.com/ambianic/ambianic-edge/blob/8c6d0d2ca22e5c72f30dd5e58324d162f085de33/tests/pipeline/ai/test_fall_detect.py#L246): [W293](https://duckduckgo.com/?q=pep8%20W293) blank line contains whitespace
> 
> ##### Comment last updated at 2020-12-18 19:58:05 UTC

@bhavikapanara could you please clean up the code to mute these pep8 warnings.

src/ambianic/pipeline/ai/fall_detect.py
```diff
@@ -171,34 +185,41 @@ def fall_detect(self, image=None):
...
-                        log.info("Fall detected: %r", inference_result)
+                    if (left_angle_with_yaxis and self._prev_left_angle_with_yaxis and left_angle_with_yaxis > self._prev_left_angle_with_yaxis) or (rigth_angle_with_yaxis and self._prev_right_angle_with_yaxis and rigth_angle_with_yaxis > self._prev_right_angle_with_yaxis):
+                        left_angle = 0
+                        left_body_line_detected = len(pose_vals_list[0]) == 2
+                        if left_body_line_detected and (self._prev_vals[0] is not None):
+                            try:
+                                temp_left_point = [[self._prev_vals[0][0], self._prev_vals[0][1]], [pose_vals_list[0][0], pose_vals_list[0][1]]]
+                                left_angle = self.calculate_angle(temp_left_point)
+                                log.info("Left shoulder-hip angle: %r", left_angle)
+                            except IndexError as e:
```
ivelin, line 196: @bhavikapanara during my testing, I found out that IndexError is thrown when the calculate_angle method is called with incomplete lines, e.g. hip only keypoint but not shoulder. I did add code to prevent that 
```
if left_body_line_detected and (self._prev_vals[0] is not None):
```

Do we still need to guard for IndexError? If so, when would that happen?

src/ambianic/pipeline/ai/fall_detect.py
```diff
@@ -171,34 +185,41 @@ def fall_detect(self, image=None):
...
-                            log.exception("Error while calculating angle: %r", e)
-
-                    angle_change = max(left_angle, right_angle)
-                    if angle_change > self._fall_factor:
-                        # insert a box that covers the whole image as a workaround
-                        # to meet the expected format of the save_detections element
-                        box = [0, 0, 1, 1]
-                        inference_result.append(('FALL', pose.score, box, angle_change ))
-                        log.info("Fall detected: %r", inference_result)
+                    if (left_angle_with_yaxis and self._prev_left_angle_with_yaxis and left_angle_with_yaxis > self._prev_left_angle_with_yaxis) or (rigth_angle_with_yaxis and self._prev_right_angle_with_yaxis and rigth_angle_with_yaxis > self._prev_right_angle_with_yaxis):
```
ivelin, line 188: @bhavikapanara I think this logic is correct, but as a best practice, we should avoid such long and convoluted conditionals.

Please extract this conditional into a separate method (maybe is_body_line_motion_downward() ) and document its purpose for better code maintainability. Thank you!

ivelin: We will respectively need to add a few test cases to cover the various scenarious.

src/ambianic/pipeline/ai/fall_detect.py
```diff
@@ -171,34 +185,41 @@ def fall_detect(self, image=None):
...
+                        left_angle = 0
```
ivelin, line 189: @bhavikapanara I think your new code for calculating body line angle with y-axis needs to be guarded with these checks. Otherwise we run the risk of exceptions in cases when we don't have a full body line detected, but only a single point (hip only or shoulder only).

src/ambianic/pipeline/ai/fall_detect.py
```diff
@@ -171,34 +185,41 @@ def fall_detect(self, image=None):
...
+                        left_body_line_detected = len(pose_vals_list[0]) == 2
+                        if left_body_line_detected and (self._prev_vals[0] is not None):
+                            try:
+                                temp_left_point = [[self._prev_vals[0][0], self._prev_vals[0][1]], [pose_vals_list[0][0], pose_vals_list[0][1]]]
+                                left_angle = self.calculate_angle(temp_left_point)
+                                log.info("Left shoulder-hip angle: %r", left_angle)
+                            except IndexError as e:
+                                log.exception("Error while calculating angle: %r", e)
+                        right_angle = 0
+                        right_body_line_detected = len(pose_vals_list[1]) == 2
```
ivelin, line 199: @bhavikapanara since this method is becoming too long and complex, I suggest breaking out redundant code for left and ride side into a separate method that is side agnostic.

src/ambianic/pipeline/ai/fall_detect.py
```diff
@@ -88,7 +90,17 @@ def calculate_angle(self, p):
         theta = abs(theta1-theta2)
         return theta
 
-
+    def calculate_angle_with_yaxis(self, p):
+        
+        x3, y3 = p[0][0], p[0][1]
+        x4, y4 = p[1][0], p[1][1]
+        
+        theta1 = 90
```
ivelin, line 98: @bhavikapanara have you verified which quadrant is the 90' vector? I've noticed that the 90' rotation with PIL.transpose() places the image in a quadrant below the X-axis, as opposed to above the X-axis as I expected. This probably has to do with the reverse Y-axis coordinates in image pixels.

Can you please visualize test_case_5 in a colab so we can be sure the vectors are correct.

It may be safer to use a vector y_axis=((0, image.height), (0, 0)) and apply the existing method:
```
calculate_angle(body_line, y-axis).
```

Think about it and let me know your thoughts.

ivelin (changes requested):
Please feel free to use the attached images for test cases. They are from actual detections with ambianic users.

Screen Shot 2020-12-18 at 4 05 30 PM
Screen Shot 2020-12-18 at 4 05 15 PM
Screen Shot 2020-12-18 at 4 04 58 PM
Screen Shot 2020-12-18 at 4 04 35 PM
Screen Shot 2020-12-18 at 4 02 25 PM
Screen Shot 2020-12-18 at 4 02 18 PM

src/ambianic/pipeline/ai/fall_detect.py
bhavikapanara: @ivelin In this case, we also need to check that **`prev_val`** have both shoulder and hip points in order to find an angle.

`left_body_line_detected = len(pose_vals_list[0]) == len(self._prev_vals[0]) == 2`
`right_body_line_detected = len(pose_vals_list[1]) == len(self._prev_vals[1]) == 2`

```
if left_body_line_detected:
     temp_left_point = [[self._prev_vals[0][0], self._prev_vals[0][1]], [pose_vals_list[0][0], pose_vals_list[0][1]]]
     left_angle = self.calculate_angle(temp_left_point)
     log.info("Left shoulder-hip angle: %r", left_angle)
```

what's your thought?

src/ambianic/pipeline/ai/fall_detect.py
bhavikapanara: > @bhavikapanara I think this logic is correct, but as a best practice, we should avoid such long and convoluted conditionals.
> 
> Please extract this conditional into a separate method (maybe is_body_line_motion_downward() ) and document its purpose for better code maintainability. Thank you!

Done

src/ambianic/pipeline/ai/fall_detect.py
bhavikapanara: @ivelin updated the code

src/ambianic/pipeline/ai/fall_detect.py
bhavikapanara: Yes, @ivelin 

You are correct...I have updated the code.

src/ambianic/pipeline/ai/fall_detect.py
bhavikapanara: @ivelin I have tried to clean up the code. Please take a look.

src/ambianic/pipeline/ai/fall_detect.py
bhavikapanara: > We will respectively need to add a few test cases to cover the various scenarious.

working to get images for test cases to cover this scenario.

src/ambianic/pipeline/ai/fall_detect.py
ivelin: @bhavikapanara did you see this comment about setting the `confidence_threshold` config parameter value.

src/ambianic/pipeline/ai/fall_detect.py
```diff
@@ -76,22 +79,42 @@ def process_sample(self, **sample):
 
 
     def calculate_angle(self, p):
```
ivelin, line 81: @bhavikapanara as I requested earlier, please add some minimal comments about the purpose of a method.

src/ambianic/pipeline/ai/fall_detect.py
```diff
@@ -76,22 +79,42 @@ def process_sample(self, **sample):
...
-        x1, y1 = p[0][0][0], p[0][0][1]
-        x2, y2 = p[0][1][0], p[0][1][1]
-        x3, y3 = p[1][0][0], p[1][0][1]
-        x4, y4 = p[1][1][0], p[1][1][1]
+
+        angle = 0
+
+        body_line_detected = len(p[0]) == len(p[1]) == 2
```
ivelin, line 85: @bhavikapanara please add a comment why we want this expression to be true. What does it accomplish in layman terms?
Since you are using non-descriptive variable names such as p[], the code is hard to read and maintain by others on the team without some commentary.

src/ambianic/pipeline/ai/fall_detect.py
```diff
@@ -76,22 +79,42 @@ def process_sample(self, **sample):
...
         theta1 = math.degrees(math.atan2(y2-y1, x1-x2))
         theta2 = math.degrees(math.atan2(y4-y3, x3-x4))
             
-        theta = abs(theta1-theta2)
-        return theta
+        angle = abs(theta1-theta2)
+        return angle
+
+
+    def is_body_line_motion_downward(self, left_angle_with_yaxis, rigth_angle_with_yaxis):
```
ivelin, line 102: @bhavikapanara this is a nice , short, clean and descriptive method. Does not need extra comments. Nice reference point for other methods.

src/ambianic/pipeline/ai/fall_detect.py
```diff
@@ -159,6 +180,11 @@ def fall_detect(self, image=None):
                     elif label == 'right shoulder' or label == 'right hip':
```
ivelin, line 180: @bhavikapanara I noticed this earlier, but forgot to mention. Since the order of points is important in relation to vector arithmetics, we should make sure that pose_vals_list[i][0] is always shoulder and pose_vals_list[i][1] is always hip. 
Since the or statement does not guarantee that, is it possible that we will add them to the list in reverse order [hip, shoulder] which will reverse the direction of the vector and the effective detection will be for someone standing up instead of falling. 

Instead of iterating over each keypoint items list and hoping they come always in the order we expect, should we optimize it to just use a lookup for 'left shoulder', 'left hip', 'right shoulder' and 'right hip'?

That would not only remove the possibility of iterating out of order but it will also remove the cycles iterating over keypoints we don't use.

Thoughts?

src/ambianic/pipeline/ai/fall_detect.py
```diff
@@ -169,47 +195,33 @@ def fall_detect(self, image=None):
...
-                    right_angle = 0
-                    right_body_line_detected = len(pose_vals_list[1]) == 2
-                    if right_body_line_detected and (self._prev_vals[1] is not None):
-                        try:
-                            temp_right_point = [[self._prev_vals[1][0], self._prev_vals[1][1]], [pose_vals_list[1][0], pose_vals_list[1][1]]]
-                            right_angle = self.calculate_angle(temp_right_point)
-                            log.info("Right shoulder-hip angle: %r", left_angle)
-                        except IndexError as e:
-                            log.exception("Error while calculating angle: %r", e)
+                    temp_left_point = [self._prev_vals[0], pose_vals_list[0]]
```
ivelin, line 201: Is `temp_left_point` the correct choice of variable name? Should it be `temp_left_vector` since it contains two points for the body shoulder-hip vector?

ivelin (changes requested):
Thank you for the update @bhavikapanara . Please see a few new clean up comments.

src/ambianic/pipeline/ai/fall_detect.py
bhavikapanara: changed...

src/ambianic/pipeline/ai/fall_detect.py
bhavikapanara: thanks @ivelin

src/ambianic/pipeline/ai/fall_detect.py
ivelin: @bhavikapanara thoughts ^^^?

src/ambianic/pipeline/ai/fall_detect.py
ivelin: I don't see a commit.

src/ambianic/pipeline/ai/fall_detect.py
bhavikapanara: @ivelin I have made changes to tackle the problem of `"make sure that pose_vals_list[i][0] is always shoulder and pose_vals_list[i][1] is always hip."`

src/ambianic/pipeline/ai/fall_detect.py
bhavikapanara: Done

src/ambianic/pipeline/ai/fall_detect.py
bhavikapanara: Updated

src/ambianic/pipeline/ai/fall_detect.py
bhavikapanara: @ivelin I didn't get exactly here what I need to do?

ivelin: The essence of the issue is that your code will bomb if `confidence_threshold` is not provided explicitly in the config file. Think about setting a good default value if no explicit value is provided by the user.

src/ambianic/pipeline/ai/fall_detect.py
ivelin: > working to get images for test cases to cover this scenario.

OK, please use the new images contributed by our users.

src/ambianic/pipeline/ai/fall_detect.py
ivelin: It's good, I like the explicit self-documenting expression. 

Although in the existing code`len(pose_vals_list[0]) == 2` guards and assures that `len(self._prev_vals[0]) == 2`, otherwise `self._prev_vals[0]` would not be assigned any value.

src/ambianic/pipeline/ai/fall_detect.py
ivelin: @bhavikapanara Looks like this is the last open comment for this PR. All others have been resolved.

src/ambianic/pipeline/ai/fall_detect.py
```diff
@@ -39,7 +39,7 @@ def __init__(self,
 
         self._pose_engine = PoseEngine(self._tfengine)
         self._fall_factor = 60
-        self.pose_confidence_threshold = kwargs['confidence_threshold']
+        self.pose_confidence_threshold = confidence_threshold
```
ivelin, line 42: @bhavikapanara you addressed the gist of the issue. Thank you!

A couple of notes:
1. As a best practice, every time we identify a potential weakness in the code and fix it, we should also add test coverage to ensure that the problem is solved correctly and to prevent future regression.
2. This solution introduces other issues that I brought up in the original comment.
Namely that `confidence_threshold` is an object instance attribute already used in the super class.
However since the super-class assumes certain properties for image detection, it is a good time to refactor it.

I will accept this PR and merge and will then open a separate issue to refactor FallDetector and TFImageDetection.

ivelin (approved):

Issue #50, opened 2019-10-03
Push notification to user


ivelin:
- [ ] notify web app (PWA Push Notification) that new samples are available to show to users.
- [ ] notify external notification handler (e.g. IFTTT, Zapier, MQTT) that allows integration with a wide variety of end points.

ivelin:
Related resources:

https://www.pushpro.io/blog/pwa-push-notifications-for-progressive-web-apps

https://platform.ifttt.com/docs#3-describe-your-queries-triggers-and-actions

ivelin:
@muka posted a few relevant resources above as I'm digging deeper into the notifications feature design.

Current thinking is to introduce a `trigger_notification` parameter to the `save_detections` pipeline element. It will have three possible values: always, positive_only, negative_only. Meaning that a notification will be trigger every time a detection event is saved, or only when a positive event is detected or only when a negative is detected (e.g. when a positive detection is no longer ongoing).

I also thought about having a separate notification element, but since  a notification should be just the minimum information without revealing any privacy details, the user will have to be able to pull details from the edge device, which requires the saved event data.

Thoughts?

ivelin:
@muka  The shoutrrr project looks relevant. From the watchdog author. https://github.com/containrrr/shoutrrr

muka:
Great finding! Will look into those libraries. Planning to work over it on the next days

PR #276, opened 2020-12-30
feature: added notification

muka:
Initial implementation for the notification system. Refers to #50

### DIFF
--- a/config.local.yaml
+++ b/config.local.yaml
@@ -10,6 +10,17 @@ default:
   timeline:
     event_log: ./data/timeline-event-log.yaml
 
+  notifications:
+    providers: {}
+    template:
+      title: "[Ambianic.ai] {event} recognized"
+      body: >
+        {event} event has been recognized by the system.
+        Visit https://ui.ambianic.ai/ for more details.
+    labels:
+      fall_detection: Fall
+      object_detection: Object
+
   sources: {}
 
   ai_models:
--- a/config.local.yaml
+++ b/config.local.yaml
@@ -17,9 +17,6 @@ default:
       body: >
         ${event} event has been recognized by the system.
         Visit ${event_details_url} for more details.
-    labels:
-      fall_detection: Fall
-      object_detection: Object
 
   sources: {}
 
--- a/config.local.yaml
+++ b/config.local.yaml
@@ -13,9 +13,9 @@ default:
   notifications:
     providers: {}
     template:
-      title: "[Ambianic.ai] {event} recognized"
+      title: "[Ambianic.ai] ${event} recognized"
       body: >
-        {event} event has been recognized by the system.
+        ${event} event has been recognized by the system.
         Visit https://ui.ambianic.ai/ for more details.
     labels:
       fall_detection: Fall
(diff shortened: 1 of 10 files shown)


muka:
Maybe not 100% ready yet but time for a first round of review :)

src/ambianic/__init__.py
```diff
@@ -27,7 +27,12 @@ def __merge_secrets(config:Union[Dynaconf, DynaBox], src_config:Dynaconf = None)
         if isinstance(val, dict):
             __merge_secrets(val, src_config)
             continue
-        if isinstance(val, str) and val[0:2] == "${":
```
ivelin, line 30: @muka I am not sure why you are parsing for ${} expressions during config load. [DynaConf seems to have](https://dynaconf.readthedocs.io/en/docs_223/guides/environment_variables.html#merging-new-data-to-existing-variables) a flexible way to do that for environment and runtime variables.

If you are thinking about populating a notification message from its template string, then that should probably happen right before the message is sent. And for that purpose, [string.format() seems appropriate](https://www.w3schools.com/python/ref_string_format.asp):

```
notification_message=notification_template.format(**template_vars))
```

src/ambianic/notification.py
```diff
@@ -0,0 +1,96 @@
...
+            title = templates.get("title", "[Ambianic.ai] New ${event} event" )
+
+        message = notification.message
+        if message is None:
+            message = templates.get(
+                "message", "New ${event} recognized"
+            )
+
+        attachments = []
+        for a in notification.attach:
```
ivelin, line 58: @muka I'm not sure we need to allow attachments. Since we are touting privacy, the messages would usually have the minimal information needed for a user to get an idea what's happening. They can pull details from their device if needed. Sending attachments through third party service providers can be a privacy issue.

src/ambianic/notification.py
```diff
@@ -0,0 +1,96 @@
...
+            attachments.append(a)
+
+        template_args = {
+            "event_type": notification.event,
+            "event": labels.get(notification.event, notification.event),
+        }
+        template_args = {**template_args, **notification.data}
+
+        for key, value in template_args.items():
+            k = "${%s}" % (str(key))
```
ivelin, line 71: As I mentioned in a previous comment, string.format() might be simpler solution.

ivelin (commented):
Looks pretty good! See a few comments.

src/ambianic/__init__.py
muka: This is specifically crafted to apply secrets replacement.

src/ambianic/notification.py
muka: Yes good point, anyway i ended up convinced should be an user decision how much privacy to keep. Consider also the API could be used to notify other information, such as log or system data

src/ambianic/notification.py
muka: I choose this approach because It would fail in case a field is not found

config.yaml
```diff
@@ -11,6 +11,18 @@ logging:
   file: ./data/ambianic-log.txt
   level: INFO
 
+# Store notifications provider configuration
+# see https://github.com/caronc/apprise#popular-notification-services for syntax examples
+# notifications:
+#   email:
```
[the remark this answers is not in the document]
muka: I do agree in part because we would loose flexibility. Currently we focus on a use case but users could appreciate this on different pipelines. Automation task could benefit of this approach for example. Can we try this approach and validate with end users first?

config.local.yaml
```diff
@@ -10,19 +10,35 @@ default:
   timeline:
     event_log: ./data/timeline-event-log.yaml
 
+  notifications:
+    providers: {}
+    template:
+      title: "[Ambianic.ai] ${event} recognized"
+      body: >
+        ${event} event has been recognized by the system.
+        Visit https://ui.ambianic.ai/ for more details.
```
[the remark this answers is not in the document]
muka: Done!

config.local.yaml
```diff
@@ -10,19 +10,35 @@ default:
...
+    labels:
+      fall_detection: Fall
```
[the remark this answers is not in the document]
muka: Done!

config.yaml
```diff
@@ -77,5 +89,8 @@ pipelines:
      - save_detections: # save samples from the inference results
         positive_interval: 10
         idle_interval: 600000
+#        notify: # notify a thirdy party service
```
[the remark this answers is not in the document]
muka: The idea was to define a set of providers and select then per element which to trigger. So, object detection may trigger a different hook than fall detection.

config.yaml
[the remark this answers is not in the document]
ivelin: Sounds good.

config.yaml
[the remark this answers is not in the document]
ivelin: Yes, I understand. Just wanted to make it simpler for you and users, but if you prefer to proceed with your original design, I'm ok with it.

src/ambianic/__init__.py
ivelin: I see. Where do the secrets values come from? Not environment? Is there a doc reference how users can use this?

src/ambianic/notification.py
ivelin: OK, got it.

config.yaml
ivelin, line 17: If we are going to leave in this flexibility for defining notification targets, can we use as examples more intuitive labels. 
`email` and `alert_fall` do not seem like an intuitive siblings under notifications. Maybe`catch_all_email` and `alert_fall` instead?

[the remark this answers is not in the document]
muka: Great thanks!

muka: Ok, done!

ivelin:
@muka LGTM. I have no further comments. Do you want me to merge now or have more to do on this PR?

muka:
We are good to go for me, thanks

ivelin:
Alright. Happy New 2021 PR merge!

Issue #277, opened 2021-01-01
[ENHANCEMENT] Refactor FallDetector and TFImageDetection


ivelin:
**Is your feature request related to a problem? Please describe.**
Currently FallDetector inherits from TFImageDetection, [which introduces several maintainability issues](https://github.com/ambianic/ambianic-edge/pull/270#discussion_r550797916):
1. `confidence_threshold`is initialized in FallDetector as a new instance attribute, even though it is expected in the TFImageDetection constructor and used in the super class as an instance attribute. We need to clean up this redundancy, because it makes the code harder to maintain.
2. TFImageDetection includes methods with assumptions about the underlying ML model that do not hold true for PoseNet. For example it expects that the model output tensor has a bounding box. 

**Describe the solution you'd like**
We need to take these specific assumptions out and construct a more generic abstract class without them - maybe TFDetectionModel. TFImageDetection and FallDetector can separately inherit from TFDetectionModel. At that point TFImageDetection should be probably renamed to TFBoundingBoxDetection to better represent its purpose.

**Describe alternatives you've considered**
Open to alternative suggestions that would help keep the code clean and maintainable. Along these lines, we will move FallDetection and other future custom models towards their own independent repos with own packaging and pypi distribution.
 
**Additional context**
https://github.com/ambianic/fall-detection
https://github.com/ambianic/fall-detection/issues/1

ivelin:
Closed by #278

PR #278, opened 2021-01-08
fix: fall detection (update store.py for fall detect)

bhavikapanara:
update the code to work on this [issue](https://github.com/ambianic/ambianic-edge/issues/277)

### DIFF
--- a/src/ambianic/pipeline/ai/fall_detect.py
+++ b/src/ambianic/pipeline/ai/fall_detect.py
@@ -266,8 +266,9 @@ class FallDetector(TFImageDetection):
                     if angle_change > self._fall_factor:
                         # insert a box that covers the whole image as a workaround
                         # to meet the expected format of the save_detections element
-                        box = [0, 0, 1, 1]
-                        inference_result.append(('FALL', pose.score, box, angle_change))
+                        #box = [0, 0, 1, 1]
+                        #inference_result.append(('FALL', pose.score, box, angle_change))
+                        inference_result.append(('FALL', pose.score, angle_change))
                         log.info("Fall detected: %r", inference_result)
 
                 log.info("Saving pose for subsequent comparison.")
--- a/src/ambianic/pipeline/ai/fall_detect.py
+++ b/src/ambianic/pipeline/ai/fall_detect.py
@@ -1,5 +1,5 @@
 """Fall detection pipe element."""
-from ambianic.pipeline.ai.image_detection import TFImageDetection
+from ambianic.pipeline.ai.tf_detect import TFDetectionModel
 from ambianic.pipeline.ai.pose_engine import PoseEngine
 from ambianic.util import stacktrace
 import logging
@@ -10,10 +10,10 @@ from PIL import Image, ImageDraw
 log = logging.getLogger(__name__)
 
 
-class FallDetector(TFImageDetection):
+class FallDetector(TFDetectionModel):
     """Detects falls comparing two images spaced about 1-2 seconds apart."""
     def __init__(self,
-                 model=None, confidence_threshold = 0.25, 
+                 model=None, 
                  **kwargs
                  ):
         """Initialize detector with config parameters.
@@ -39,7 +39,7 @@ class FallDetector(TFImageDetection):
 
         self._pose_engine = PoseEngine(self._tfengine)
         self._fall_factor = 60
-        self.pose_confidence_threshold = confidence_threshold
+        self.pose_confidence_threshold = self._tfengine._confidence_threshold
 
         # Require a minimum amount of time between two video frames in seconds.
         # Otherwise on high performing hard, the poses could be too close to each other and have negligible difference
--- a/src/ambianic/pipeline/store.py
+++ b/src/ambianic/pipeline/store.py
@@ -91,22 +91,40 @@ class SaveDetectionSamples(PipeElement):
         inf_json = []
         if inference_result:
             for inf in inference_result:
-                label, confidence, box = inf[0:3]
-                log.info('label: %s , confidence: %.0f, box: %s',
-                        label,
-                        confidence,
-                        box)
-                one_inf = {
-                    'label': label,
-                    'confidence': float(confidence),
-                    'box': {
-                        'xmin': float(box[0]),
-                        'ymin': float(box[1]),
-                        'xmax': float(box[2]),
-                        'ymax': float(box[3]),
+                label, confidence = inf[0:2]
+
+                if isinstance(inf[2], (list, tuple)):
+                    # Process inference for person/object detection model
+                    box = inf[2]
+                    log.info('label: %s , confidence: %.0f, box: %s',
+                            label,
+                            confidence,
+                            box)
+                    one_inf = {
+                        'label': label,
+                        'confidence': float(confidence),
+                        'box': {
+                            'xmin': float(box[0]),
+                            'ymin': float(box[1]),
+                            'xmax': float(box[2]),
+                            'ymax': float(box[3]),
+                        }
                     }
-                }
-                inf_json.append(one_inf)
+                    inf_json.append(one_inf)
+                else:
+                    # Process inference for Fall detection model
+                    angle_diff_posenet = inf[2]
+                    log.info('label: %s , confidence: %.0f, angle difference: %s',
+                            label,
+                            confidence,
+                            angle_diff_posenet)
+                    one_inf = {
+                        'label': label,
+                        'confidence': float(confidence),
+                        'angle_diff': angle_diff_posenet
+                    }
+                    inf_json.append(one_inf)
+
         save_json = {
             'id': uuid.uuid4().hex,
             'datetime': inf_time.isoformat(),
--- a/tests/pipeline/ai/test_fall_detect.py
+++ b/tests/pipeline/ai/test_fall_detect.py
@@ -161,8 +161,7 @@ def test_fall_detection_case_2_2():
 
     assert result
     assert len(result) == 1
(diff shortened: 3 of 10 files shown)


src/ambianic/pipeline/ai/fall_detect.py
```diff
@@ -266,8 +266,9 @@ def fall_detect(self, image=None):
                     if angle_change > self._fall_factor:
                         # insert a box that covers the whole image as a workaround
                         # to meet the expected format of the save_detections element
-                        box = [0, 0, 1, 1]
```
ivelin, line 269: @bhavikapanara simply removing the box will cause an error in the UI, but that's something we will address separately in the UI logic.

Let's add to the inference_result the pose keypoints. The intention is to store a detailed sample of a fall detection event with rich parameters that we can use in UI rendering and also in the future for model performance analysis and training.

src/ambianic/pipeline/ai/fall_detect.py
```diff
@@ -266,8 +266,9 @@ def fall_detect(self, image=None):
...
-                        inference_result.append(('FALL', pose.score, box, angle_change))
+                        #box = [0, 0, 1, 1]
+                        #inference_result.append(('FALL', pose.score, box, angle_change))
+                        inference_result.append(('FALL', pose.score, angle_change))
```
ivelin, line 271: Let's also store the before and after images along with their keypoints to help debug fall detection issues (such as false positives). We may also end up rendering in the UI before and after image details with a fall vector to make it clear to the user WHY the AI detected a fall. It would also give the user a richer context of WHY and HOW the person fell.

src/ambianic/pipeline/store.py
```diff
@@ -91,22 +91,40 @@ def _save_sample(self,
...
-                    'label': label,
-                    'confidence': float(confidence),
-                    'box': {
-                        'xmin': float(box[0]),
-                        'ymin': float(box[1]),
-                        'xmax': float(box[2]),
-                        'ymax': float(box[3]),
+                label, confidence = inf[0:2]
+
+                if isinstance(inf[2], (list, tuple)):
```
ivelin, line 96: It think it's time to introduce more structure and easier to maintain logic. Remember the earlier issue to refactor TFImageDetection? 

store.py needs to use a more intuitive logic for saving various detections. Maybe we pull the code for generating a serializable event structure in the detection classes and store simply uses that and dumps it to the event log. 

As we are adding more detections, it becomes unclear why store needs to know the internal data structure details of each detection output.

tests/pipeline/ai/test_fall_detect.py
```diff
@@ -161,8 +161,7 @@ def sample_callback(image=None, inference_result=None, **kwargs):
 
     assert result
     assert len(result) == 1
-    category, confidence, box, angle = result[0]
-    assert box   # Add this line to avoid 'Unused local variable' 
+    category, confidence, angle = result[0]
```
ivelin, line 164: @bhavikapanara Simply removing the box detection makes this test pass but breaks the UI. We need to discuss an event data format that the edge and UI can agree on. Until fall detection, all inferences had the same output format. Now that we are introducing keypoints and angles as opposed to boxes, we need to rethink how the edge and UI agree on detection format.

tests/pipeline/ai/test_fall_detect.py
```diff
@@ -192,8 +191,7 @@ def sample_callback(image=None, inference_result=None, **kwargs):
...
```
ivelin, line 194: Let's add test conditions for all fall detection parameters: keypoints for example. What else are we missing to verify in this test?

src/ambianic/pipeline/store.py
```diff
@@ -91,22 +91,40 @@ def _save_sample(self,
...
+                    # Process inference for person/object detection model
+                    box = inf[2]
```
bhavikapanara, line 98: @ivelin I have written a code to manage inference type whether it contains bounding box or angle difference. As I am not able to see changes in UI, it may fail in UI.

src/ambianic/pipeline/ai/fall_detect.py
bhavikapanara: @ivelin I have written a code to manage inference type whether it contains bounding box or angle difference in store.py file. As I am not able to see changes in UI, it may fail in UI.

ivelin (changes requested):
Two main issues here:

1. We need to discuss and define a new data format that the edge and UI can agree on for passing detection event information. @muka your input would be welcome here.
2. Fall detection tests should cover all key parameters, such as keypoints. Look at the test coverage report and see what other code logic we may not be testing for.

src/ambianic/pipeline/ai/fall_detect.py
bhavikapanara: @ivelin do you want to add all 17 pose keypoints to inference_result? At this moment inference_result contain angle difference.

tests/pipeline/ai/test_fall_detect.py
bhavikapanara: @ivelin ok...then let's first finalize output format. Next, will modify the code accordingly.

tests/pipeline/ai/test_fall_detect.py
ivelin: @bhavikapanara did you see my comments to this PR?

tests/pipeline/ai/test_fall_detect.py
ivelin: @bhavikapanara Here is how we will proceed here. Instead of redefining the event format, we will populate a bounding box value by calculating the box surrounding all detected keypoins. See [this code](https://github.com/tensorflow/tfjs-models/blob/08dd2ed9ffcd3f3604026c2374f99a3498184d33/posenet/src/util.ts#L45) from the posenet js demo.

```
const {NEGATIVE_INFINITY, POSITIVE_INFINITY} = Number;
export function getBoundingBox(keypoints: Keypoint[]):
    {maxX: number, maxY: number, minX: number, minY: number} {
  return keypoints.reduce(({maxX, maxY, minX, minY}, {position: {x, y}}) => {
    return {
      maxX: Math.max(maxX, x),
      maxY: Math.max(maxY, y),
      minX: Math.min(minX, x),
      minY: Math.min(minY, y)
    };
  }, {
    maxX: NEGATIVE_INFINITY,
    maxY: NEGATIVE_INFINITY,
    minX: POSITIVE_INFINITY,
    minY: POSITIVE_INFINITY
  });
}
```

This will resolve the immediate issue with format incompatibility in a meaningful way.

We should still add keypoints as an additional event attribute, which will be logged to disk and used by the UI to render overlay of a pose skeleton.

src/ambianic/pipeline/store.py
ivelin: @bhavikapanara did you see this comment? It is still valid. The way we feed a bounding box to the store class should be pulled out of here and into the model output since each model has its own way of defining a bounding box.

src/ambianic/pipeline/ai/fall_detect.py
ivelin: @bhavikapanara fyi ^^^ Expecting your thoughts.

src/ambianic/pipeline/store.py
ivelin: @bhavikapanara ok, see my latest comments on how we will solve the UI interface issue. You don't need to spend time figuring out the UI side.

tests/pipeline/ai/test_fall_detect.py
bhavikapanara: @ivelin `Instead of redefining the event format, we will populate a bounding box value by calculating the box surrounding all detected keypoins.`

Didn't get it exactly what I need to do here by calculating the box surrounding all detected keypoitns.?

src/ambianic/pipeline/ai/fall_detect.py
bhavikapanara: @ivelin 
Ok...can you please list down all items that you want to add in `inference_result`?
Will solve the rest of the comments that are related to `inference_result` once I have a format of `inference_result`

ivelin: Ok, this seems more complicated than it needs to be. Let's keep things simple.
Populate the bounding box as described and do not add more data attributes in this PR. We will handle that in a separate issue and PR.
For output score value, use the average of the hip&shoulder scores that were actually used for angle calculation. It's a reasonable score metric to work with for now until we know that something better is needed.

src/ambianic/pipeline/ai/fall_detect.py
ivelin: @bhavikapanara yes, let's add all keypoints that are above the required keypoint threshold. The UI can decide later how to render them. Please do it in a separate PR to keep this PR scope manageable.

tests/pipeline/ai/test_fall_detect.py
ivelin: Proceed with a reasonable implementation of a bounding box. That will show a meaningful output in the current UI. A later version of the UI will also use the inference keypoints to draw a stick figure.

bhavikapanara:
> Two main issues here:
> 
> 1. We need to discuss and define a new data format that the edge and UI can agree on for passing detection event information. @muka your input would be welcome here.
> 2. Fall detection tests should cover all key parameters, such as keypoints. Look at the test coverage report and see what other code logic we may not be testing for.

@ivelin 

Can you please finalize a data format for fall detection?

Issue #279, opened 2021-01-12
Improve fall detection with new confidence_threshold formula


ivelin:
**Is your feature request related to a problem? Please describe.**
Currently in the fall detection algorithm, confidence_thresold [applies to the average of all 17 keypoint scores](https://github.com/ambianic/ambianic-edge/blob/c418f4db5e6487aa086579ade56347e35ad12c83/src/ambianic/pipeline/ai/pose_engine.py#L166) from a pose detection model output. Upon further experimentation and user testing, it is becoming clear that this approach is sub-optimal.

In practical situations, the user whose fall we intend to detect may not be facing the camera. Or parts of the user body may be obstructed by furniture, doors or other household items.

By limiting detection to an average of all body keypoints, we cut out a number of practical situations.

**Describe the solution you'd like**
We should consider high confidence score only for the parts that are relevant to detecting a fall. In the current algorithm we use the hip-shoulder vectors (left or right). As long as there is a high level of confidence that the model detected any one of the body torso vectors, we should feed that vector to the [downward motion heuristic](https://github.com/ambianic/ambianic-edge/blob/c418f4db5e6487aa086579ade56347e35ad12c83/src/ambianic/pipeline/ai/fall_detect.py#L105).

We should end up with something like this:
- extract all hip and shoulder keypoints along with their score
- leftVectorScore = average (left hip score + left shoulder score)
- if leftVectorScore > confidence_threshold then use it in calculating angles. Otherwise skip it in this detection run.
- rightVectorScore  = average (right hip score + right shoulder score)
- if rightVectorScore > confidence_threshold then use it in calculating angles. Otherwise skip it in this detection run.
- lean_angle = get the max of the leaning angles for all of the vectors that passes the confidence_threshold test
- if lean_angle > lean_threshold, then indicate a positive fall detection

**Additional context**
We will still have to test if this change results in false positives, which we may be able to mitigate with a higher confidence_threshold.

bhavikapanara:
@ivelin 

In PR #280, I have added the following conditions:

```
extract all hip and shoulder keypoints along with their score
leftVectorScore = average (left hip score + left shoulder score)
if leftVectorScore > confidence_threshold then use it in calculating angles. Otherwise skip it in this detection run.
rightVectorScore = average (right hip score + right shoulder score)
if rightVectorScore > confidence_threshold then use it in calculating angles. Otherwise skip it in this detection run.
[...]

PR #280, opened 2021-01-12
fix:  fall detection issues #279

bhavikapanara:


src/ambianic/pipeline/ai/fall_detect.py
```diff
@@ -243,11 +243,18 @@ def fall_detect(self, image=None):
                 pose_dix = {}
                 inference_result = []
 
-                for e_corr in self.fall_detect_corr:
-                    if pose.keypoints[e_corr].score > 0.6:
-                        # detected key-point probability must be greater than 0.6
-                        pose_dix[e_corr] = pose.keypoints[e_corr].yx
-                                        
+                # Find leftVectorScore and rightVectorScore --> i.e average score of shoulder & hip
+                leftVectorScore = (pose.keypoints['left shoulder'].score + pose.keypoints['left hip'].score) / 2.0
```
ivelin, line 247: @bhavikapanara I apologize but I just realized that the average of the hip and shoulder scores is not optimal. What if the hip or shoulder are out of the picture? I've seen users show me images where the model detects a shoulder point correctly, but the hip is some random pixel that is not at all on the body. 

I think we need to apply the confidence threshold to each of the individual hip/shoulder key points. If both hip and shoulder on one side pass the confidence threshold, then we can use the vector, otherwise ignore it.

Making sense?

ivelin: I am attaching a few images that you can use in unit tests to ensure that confidence threshold is used in a meaningful way. I checked the original posenet model demo code (JavaScript) and noticed that it uses two separate parameters: one for the minimal score required for each individual keypoint and another score for the overall average across all keypoints. I think for our case we may be able to get away with applying the confidence threshold (score) to each individual keypoint that represents a hip or shoulder. Ignore other keypoints. Once its implemented we will do more testing with users to verify if that's a stable predictor.

ivelin: Screen Shot 2021-01-12 at 2 10 23 PM
Screen Shot 2021-01-12 at 2 10 10 PM

src/ambianic/pipeline/ai/image_boundingBox_detection.py
```diff
@@ -7,12 +7,12 @@
...
 from .inference import TFInferenceEngine
 from ambianic.pipeline import PipeElement
-
+from ambianic.pipeline.ai.tf_detect import TFDetectionModel
 
 log = logging.getLogger(__name__)
 
 
-class TFImageDetection(PipeElement):
+class TFBoundingBoxDetection(PipeElement):
```
ivelin, line 15: @bhavikapanara Should TFBoundingBoxDetection extend TFDetectionModel? 
It seems like __ini__, load_labels, and log_stats are reusable.

ivelin (commented):
@bhavikapanara could you please review the pep8speaks alerts?

src/ambianic/pipeline/ai/tf_detect.py
```diff
@@ -0,0 +1,180 @@
...
+        try:
+            # convert from numpy to native Python int type
+            # that PIL expects
+            if isinstance(w, np.generic):
+                w = w.item()
+                w = int(w)
+                h = h.item()
+                h = int(h)
+            thumb.thumbnail((w, h))
+        except Exception as e:
```
ivelin, line 94: @bhavikapanara could you please add unit test coverage for the lines the codecov flagged?

ivelin (changes requested):
Thank you for the PR. Good work. It addresses much of the refactoring issue. I made a few comments that require your input.

Please keep this PR scoped to only the refactoring. Use separate PRs for the other open issues.

Thank you!

ivelin:
@bhavikapanara please use the user provided images that I uploaded to this PR for regression tests. Otherwise we may be making one step forward and two back.

bhavikapanara:
> @bhavikapanara could you please review the pep8speaks alerts?

src/ambianic/pipeline/ai/fall_detect.py
bhavikapanara: @ivelin  Can you please share only original person image, not screenshot?
I need to feed similar kind of images as we have already used for other tests like
fall_img_1

ivelin:
I don’t have the original, but you can crop it with an editor. The image
size is the thumbnail which is bigger than or equal to the input tensor so
inference should work fine.

src/ambianic/pipeline/ai/fall_detect.py
bhavikapanara: > @bhavikapanara I apologize but I just realized that the average of the hip and shoulder scores is not optimal. What if the hip or shoulder are out of the picture? I've seen users show me images where the model detects a shoulder point correctly, but the hip is some random pixel that is not at all on the body.
> 
> I think we need to apply the confidence threshold to each of the individual hip/shoulder key points. If both hip and shoulder on one side pass the confidence threshold, then we can use the vector, otherwise ignore it.
> 
> Making sense?

So, in that case, our [previous logic](https://github.com/ambianic/ambianic-edge/blob/c418f4db5e6487aa086579ade56347e35ad12c83/src/ambianic/pipeline/ai/fall_detect.py#L247) is correct to check the score individually for hip and shoulder.

One solution can be to consider only hip and shoulder points while calculating the pose_score as we are more interested only in 4 points left and right shoulders and hips instead of t[aking the average of all 17 key-points.](https://github.com/ambianic/ambianic-edge/blob/c418f4db5e6487aa086579ade56347e35ad12c83/src/ambianic/pipeline/ai/pose_engine.py#L166)

For example:
Proposed Solution

```
fall_detect_corr = ['left shoulder', 'left hip', 'right shoulder', 'right hip']

for point_i in range(kps.shape[0]):
    x, y = kps[point_i, 1], kps[point_i, 0]
    prob = self.sigmoid(kps[point_i, 3])

    if KEYPOINTS[point_i] in fall_detect_corr and  prob > 0.60:
        cnt += 1

    keypoint = Keypoint(KEYPOINTS[point_i], [x, y], prob)            
    keypoint_dict[KEYPOINTS[point_i]] = keypoint
    # draw on image and save it for debugging
    draw = ImageDraw.Draw(template_image)
    draw.line(((0,0), (x, y)), fill='red')

pose_scores = cnt/4

```
Instead of 

```
for point_i in range(kps.shape[0]):
    x, y = kps[point_i, 1], kps[point_i, 0]
    prob = self.sigmoid(kps[point_i, 3])

    if prob > 0.60:
        cnt += 1
    keypoint = Keypoint(KEYPOINTS[point_i], [x, y], prob)            
    keypoint_dict[KEYPOINTS[point_i]] = keypoint
    # draw on image and save it for debugging
    draw = ImageDraw.Draw(template_image)
    draw.line(((0,0), (x, y)), fill='red')

pose_scores = cnt/17
```

bhavikapanara: @ivelin any thoughts for the above comment?

ivelin: > @ivelin any thoughts for the above comment?

Here we go:

> So, in that case, our previous logic is correct to check the score individually for hip and shoulder.

Yes, in the sense that we need to apply confidence_threshold to individual keypoints to avoid detections in completely irrelevant locations.

> One solution can be to consider only hip and shoulder points while calculating the pose_score as we are more interested only in 4 points left and right shoulders and hips instead of taking the average of all 17 key-points.

Yes, that is exactly what I am suggesting as a next step.

> Proposed Solution

Yes, that is fine. I actually don't think `pose_score` is relevant in our current fall detection score. 

Instead we have a criteria which requires that we only use a body vector (left or right) to calculate a leaning angle if and only if its individual keypoints (both hip and shoulder) exceed the confidence_threshold (instead of hardcoded 0.6).

In other words `pose_scores` is not used as a limiting criteria for fall detection. 

Maybe a strict definition of `fall_score` can be:

```
# current_body_vector is the one we use for the current angle calculation, either left or right
current_body_vector_score = (hip_score + shoulder_score)/2
leaning_probability = 1 if leaning_angle > fall_leaning_threshold else 0
fall_score = leaning_probability * (previous_body_vector_score + current_body_vector_score)
```

As we've discussed elsewhere, `leaning_probabiliy` is a trivial binary heuristic value for now with `fall_leaning_threshold = 60'. In the future we will likely replace the heuristic rule with a DNN model that is trained to recognize which differences in pose keypoints between image frames constitutes a fall.

Thoughts?

src/ambianic/pipeline/ai/fall_detect.py
```diff
@@ -243,11 +243,18 @@ def fall_detect(self, image=None):
                 pose_dix = {}
                 inference_result = []
 
-                for e_corr in self.fall_detect_corr:
```
ivelin, line 246: @bhavikapanara As you pointed out, this filter should be restored to the extent that we want each of the hip & shoulder key points to be above the `confidence_threshold` (default value 0.6 now, not 0.25).

src/ambianic/pipeline/ai/fall_detect.py
```diff
@@ -243,11 +243,18 @@ def fall_detect(self, image=None):
...
-                    if pose.keypoints[e_corr].score > 0.6:
-                        # detected key-point probability must be greater than 0.6
-                        pose_dix[e_corr] = pose.keypoints[e_corr].yx
-                                        
+                # Find leftVectorScore and rightVectorScore --> i.e average score of shoulder & hip
+                leftVectorScore = (pose.keypoints['left shoulder'].score + pose.keypoints['left hip'].score) / 2.0
+                rightVectorScore = (pose.keypoints['right shoulder'].score + pose.keypoints['right hip'].score) / 2.0
+
+                if leftVectorScore > self.pose_confidence_threshold:
```
ivelin, line 250: This test should be replaced 

from:
```
  rightVectorScore = (pose.keypoints['right shoulder'].score + pose.keypoints['right hip'].score) / 2.0
```

to:

```
  rightVectorScore = min(pose.keypoints['right shoulder'].score, pose.keypoints['right hip'].score) 
```

same for the left vector score.

src/ambianic/pipeline/ai/fall_detect.py
ivelin: Actually, there is a more succinct way to accomplish this. See the comment below for updating the formula for `right_vector_score` and `left_vector_score`.

src/ambianic/pipeline/ai/image_boundingBox_detection.py
ivelin: @bhavikapanara did you see this comment ^^^

bhavikapanara: @ivelin Can you please check here: https://github.com/ambianic/ambianic-edge/blob/9915364eaa507f89ad17124125c7b1fb8ffbc211/src/ambianic/pipeline/ai/image_boundingBox_detection.py#L39

ivelin: Right. I see that you chose to use class delegation instead of inheritance. Based on the intended use of these classes and the redundant code in each, I am not sure that delegation is the optimal choice. Maybe you can elaborate to help me understand why inheritance is not desired.

src/ambianic/pipeline/ai/fall_detect.py
bhavikapanara: Done

src/ambianic/pipeline/ai/image_boundingBox_detection.py
bhavikapanara: @ivelin Actually I faced some error for multiple inheritances at that time. Let me show you

bhavikapanara:
```
======================================================== ERRORS =========================================================
________________________________ ERROR collecting tests/pipeline/ai/test_fall_detect.py _________________________________
tests/pipeline/ai/test_fall_detect.py:3: in <module>
    from ambianic.pipeline.ai.object_detect import ObjectDetector
src/ambianic/pipeline/ai/object_detect.py:3: in <module>
    from .image_boundingBox_detection import TFBoundingBoxDetection
src/ambianic/pipeline/ai/image_boundingBox_detection.py:15: in <module>
    class TFBoundingBoxDetection(PipeElement, TFDetectionModel):
E   TypeError: Cannot create a consistent method resolution
E   order (MRO) for bases PipeElement, TFDetectionModel
```

bhavikapanara: @ivelin for that reason I used class delegation instead of inheritance. Please help me to resolve this error.

src/ambianic/pipeline/ai/fall_detect.py
bhavikapanara: @ivelin it makes sense

src/ambianic/pipeline/ai/fall_detect.py
ivelin: Awesome. We just need a couple of test cases to cover the new formula code.

src/ambianic/pipeline/ai/image_boundingBox_detection.py
ivelin: Why do you need multiple inheritance? TFDetectionModel is a PipeElement itself. You can replace 
```
class TFBoundingBoxDetection(PipeElement, TFDetectionModel):
```

with 

```
class TFBoundingBoxDetection(TFDetectionModel):
```

src/ambianic/pipeline/ai/tf_detect.py
bhavikapanara: @ivelin I didn't get any idea what exception case should I generate for this function. Please guide me.

src/ambianic/pipeline/ai/image_boundingBox_detection.py
bhavikapanara: @ivelin Ohh...I missed this point...My apology

bhavikapanara:
> @bhavikapanara please use the user provided images that I uploaded to this PR for regression tests. Otherwise we may be making one step forward and two back.

Added [test case](https://github.com/ambianic/ambianic-edge/blob/279905c058f22337e3fe05e9e6f3b3bc559f5648/tests/pipeline/ai/test_fall_detect.py#L284) for the above images.

src/ambianic/pipeline/ai/fall_detect.py
bhavikapanara: @ivelin I think we also need to update [this part](https://github.com/ambianic/ambianic-edge/blob/279905c058f22337e3fe05e9e6f3b3bc559f5648/src/ambianic/pipeline/ai/fall_detect.py#L129) as it relies on pose score which is average of all 17 key-points.

It might possible in given image shoulder and the hip point is correctly detected but due to low pose score the image gets rotated and find key-points in the rotated image even in original image hip and shoulder are found.

What's your thoughts?

bhavikapanara: > current_body_vector is the one we use for the current angle calculation, either left or right
> leaning_probability = 1 if leaning_angle > fall_leaning_threshold else 0
> fall_score = leaning_probability * (previous_body_vector_score + current_body_vector_score) / 2

@ivelin Do I need to add this part in current PR?

ivelin: Yes. We need a reasonable value for fall detection score, that will be stored on disk and shown in the UI to the user. Think about the inference output (score, box, metadata) that is expected in store.py from inference classes.

src/ambianic/pipeline/ai/tf_detect.py
ivelin: Receiving meaningless parameters should cause an exception.

bhavikapanara: @ivelin In this function there is only two parameters that are ` w, h = desired_size` and these value w,h are coming from [TensorFlow input model parameter ](https://github.com/ambianic/ambianic-edge/blob/80e1b367c49e8ba32289b9f805364518962e2886/src/ambianic/pipeline/ai/image_boundingBox_detection.py#L113)so that can't be incorrect. Please take a look at it.

```
height = tfe.input_details[0]['shape'][1]
width = tfe.input_details[0]['shape'][2]

desired_size = (width, height)
thumbnail = TFDetectionModel.thumbnail(image=image, desired_size=desired_size)
```

src/ambianic/pipeline/ai/fall_detect.py
ivelin: > @ivelin I think we also need to update [this part](https://github.com/ambianic/ambianic-edge/blob/279905c058f22337e3fe05e9e6f3b3bc559f5648/src/ambianic/pipeline/ai/fall_detect.py#L129) as it relies on pose score which is average of all 17 key-points.
> 
> It might possible in given image shoulder and the hip point is correctly detected but due to low pose score the image gets rotated and find key-points in the rotated image even in original image hip and shoulder are found.
> 
> What's your thoughts?

@bhavikapanara  Very good catch.

Your earlier suggestion comes back to light. pose_score is still relevant.

Maybe we can calculate and return pose_score as the average of the left and right vector scores:

```
pose_score = (leftVectorScore + rightVectorScore) / 2
```

This leads me to another concern. What if we see only the lift vector in the previous image and we only see the right vector in the current image? That is still useful information, but in the current logic we would just throw it away and not register any angle change (as both left and right angles will be calculated as non-existant or 0').

Ideally we would work with a learned spinal vector estimate model, but that's a bigger conversation for developing such model or using one that gets us there. See [this recent work](https://ai.googleblog.com/2021/01/recognizing-pose-similarity-in-images.html) from Google AI that may be relevant.


What do you think about using a simple heuristic for now to estimate the spinal vector:
```
if left vector is below confidence threshold, ignore it
if right vector is below confidence threshold, ignore it
if either left or right are above confidence threshold:
  spinalVectorEstimate = average of left and right vectors
if spinalVectorEstimate available:
  calculate leaning angle based on the difference in previous and current spinalVector estimates.
  angle_heuristic_score = 0 or 1 as discussed elsewhere in this PR
  fall_detection_score = angle_heuristic_score * spinalVectorEstimateScore
else:
  no detection
  fall_detection_score = 0
```

@muka I would appreciate your take on this one ^^^

bhavikapanara: @ivelin At this moment the inference output is this: 
[`inference_result.append(('FALL', pose.score, box, angle_change))`](https://github.com/ambianic/ambianic-edge/blob/15d9285ed1ae5f767598a1f2e83dcfca8a43ed50/src/ambianic/pipeline/ai/fall_detect.py#L270)

Now, we want to update it to 
```
leaning_probability = 1 if leaning_angle > fall_leaning_threshold else 0
fall_score = leaning_probability * (previous_body_vector_score + current_body_vector_score) / 2

inference_result.append(('FALL', fall_score, box, ?))

```
Here 

- leaning_angle is [angle_changes](https://github.com/ambianic/ambianic-edge/blob/15d9285ed1ae5f767598a1f2e83dcfca8a43ed50/src/ambianic/pipeline/ai/fall_detect.py#L266) b/w shoulder-hip lines in a current and previous frame
- fall_leaning_threshold --> is [`self._fall_factor` ](https://github.com/ambianic/ambianic-edge/blob/279905c058f22337e3fe05e9e6f3b3bc559f5648/src/ambianic/pipeline/ai/fall_detect.py#L41)i.e 60 degree 
- previous_body_vector_score: How to find it from [leftVectorScore](https://github.com/ambianic/ambianic-edge/blob/279905c058f22337e3fe05e9e6f3b3bc559f5648/src/ambianic/pipeline/ai/fall_detect.py#L247) and [rightVectorScore](https://github.com/ambianic/ambianic-edge/blob/279905c058f22337e3fe05e9e6f3b3bc559f5648/src/ambianic/pipeline/ai/fall_detect.py#L248)
- Similar current_body_vector_score


Also, didn't get you for metadata?

src/ambianic/pipeline/ai/tf_detect.py
ivelin: Right. The purpose of the test is to force an exception just in case this ever happens at runtime. Imagine there is a bug in the tfmodel or it outputs crazy tensor values.

src/ambianic/pipeline/ai/fall_detect.py
ivelin: > @ivelin At this moment the inference output is this:
> [`inference_result.append(('FALL', pose.score, box, angle_change))`](https://github.com/ambianic/ambianic-edge/blob/15d9285ed1ae5f767598a1f2e83dcfca8a43ed50/src/ambianic/pipeline/ai/fall_detect.py#L270)
> 
> Now, we want to update it to
> 
> ```
> leaning_probability = 1 if leaning_angle > fall_leaning_threshold else 0
> fall_score = leaning_probability * (previous_body_vector_score + current_body_vector_score) / 2
> 
> inference_result.append(('FALL', fall_score, box, ?))
> ```

Yes.

> 
> Here
> 
> * leaning_angle is [angle_changes](https://github.com/ambianic/ambianic-edge/blob/15d9285ed1ae5f767598a1f2e83dcfca8a43ed50/src/ambianic/pipeline/ai/fall_detect.py#L266) b/w shoulder-hip lines in a current and previous frame

Yes.

> * fall_leaning_threshold --> is [`self._fall_factor` ](https://github.com/ambianic/ambianic-edge/blob/279905c058f22337e3fe05e9e6f3b3bc559f5648/src/ambianic/pipeline/ai/fall_detect.py#L41)i.e 60 degree

Yes.

> * previous_body_vector_score: How to find it from [leftVectorScore](https://github.com/ambianic/ambianic-edge/blob/279905c058f22337e3fe05e9e6f3b3bc559f5648/src/ambianic/pipeline/ai/fall_detect.py#L247) and [rightVectorScore](https://github.com/ambianic/ambianic-edge/blob/279905c058f22337e3fe05e9e6f3b3bc559f5648/src/ambianic/pipeline/ai/fall_detect.py#L248)

Here I meant the body vector score calculated for the previous image and the body vector score calculated for the current image.

> * Similar current_body_vector_score

This now became a new topic after you discovered the issue in the rotation logic. Let me know your thoughts on spine estimation and pose_score formulas.

> 
> Also, didn't get you for metadata?

That's already taken care of [here](https://github.com/ambianic/ambianic-edge/blob/15d9285ed1ae5f767598a1f2e83dcfca8a43ed50/src/ambianic/pipeline/ai/fall_detect.py#L65). It's just decorative data to display in the UI for user convenience.

ivelin: Just realized that `some_angle_correction` is not trivial, because we don't know if the image is from the front, back or side of the person since its only one side vector and we have no simple reference point. Let's skip this angle correction for now. And in the near future look closer at transfer learning a DNN model such as [Pr_VIPE](https://ai.googleblog.com/2021/01/recognizing-pose-similarity-in-images.html) for 3D pose retrieval from monocular pose 2D images.

bhavikapanara: > Here I meant the body vector score calculated for the previous image and the body vector score calculated for the current image.


@ivelin Still I didn't get which body vector you are referencing here!
how to calculate **previous_body_vector_score** and **current_body_vector_score** for previous and current image? Can you please elaborate it with mathematical representation?

ivelin: @bhavikapanara 

The latest iteration is as follows:

# see formula for spinalVectorScore above
current_body_vector_score  = pose_score = spinalVectorScore
# do angle calculation
# do fall_score calculation

# save current_body_vector_score along with keypoints as previous and compare when the next image frame comes in
previous_body_vector_score = current_body_vector_score

src/ambianic/pipeline/ai/tf_detect.py
bhavikapanara: Done

src/ambianic/pipeline/ai/fall_detect.py
bhavikapanara: @ivelin 
Here required to change the condition to manage the image [rotation section](https://github.com/ambianic/ambianic-edge/blob/279905c058f22337e3fe05e9e6f3b3bc559f5648/src/ambianic/pipeline/ai/fall_detect.py#L129) by comparing newly invented pose_score instead of pose.score

https://github.com/ambianic/ambianic-edge/blob/279905c058f22337e3fe05e9e6f3b3bc559f5648/src/ambianic/pipeline/ai/fall_detect.py#L129

will become

        while (not poses or pose_score < min_score) and rotations:

And

https://github.com/ambianic/ambianic-edge/blob/279905c058f22337e3fe05e9e6f3b3bc559f5648/src/ambianic/pipeline/ai/fall_detect.py#L137
will become

        if (pose and pose_score >= min_score):

right?

src/ambianic/pipeline/ai/fall_detect.py
```diff
@@ -126,15 +127,21 @@ def find_keypoints(self, image):
...
+        while pose_score < min_score and rotations:
           angle = rotations.pop()
           transposed = image.transpose(angle)
           # we are interested in the poses but not the rotated thumbnail
           poses, _ = self._pose_engine.DetectPosesInImage(transposed)
+          pose_score, _ = self.estimate_spinalVector_score(poses[0])
+
         if poses and poses[0]:
             pose = poses[0]
+            pose_score, _ = self.estimate_spinalVector_score(pose)
```
ivelin, line 141: @bhavikapanara why do we calculate the same pose_score twice? We already have the latest pose_score from the while loop above. We can also keep a current pose reference along the current pose_score reference and reuse both to avoid the redundant code and CPU usage.

src/ambianic/pipeline/ai/fall_detect.py
```diff
@@ -222,6 +230,51 @@ def get_line_angles_with_yaxis(self, pose_dix):
         
         return (l_angle, r_angle)
 
+    def estimate_spinalVector_score(self, pose):
```
ivelin, line 233: @bhavikapanara awesome implementation of the unified spinalVectorScore algorithm!!

src/ambianic/pipeline/ai/fall_detect.py
```diff
@@ -222,6 +230,51 @@ def get_line_angles_with_yaxis(self, pose_dix):
...
+        def find_spinalLine():
+            left_spinal_x1 = (pose_dix['left shoulder'][0] + pose_dix['right shoulder'][0]) / 2
+            left_spinal_y1 = (pose_dix['left shoulder'][1] + pose_dix['right shoulder'][1]) / 2
+            
+            right_spinal_x1 = (pose_dix['left hip'][0] + pose_dix['right hip'][0]) / 2
+            right_spinal_y1 = (pose_dix['left hip'][1] + pose_dix['right hip'][1]) / 2
+            
+            return (left_spinal_x1, left_spinal_y1), (right_spinal_x1, right_spinal_y1)
+
+        is_leftVector = leftVectorScore >= self.confidence_threshold
```
ivelin, line 257: @bhavikapanara maybe we can bump `is_leftVector` a little higher in this function to remove the redundancy of `rightVectorScore > self.confidence_threshold`. Same for the left side.

src/ambianic/pipeline/ai/fall_detect.py
```diff
@@ -222,6 +230,51 @@ def get_line_angles_with_yaxis(self, pose_dix):
...
+            spinalVectorEstimate = find_spinalLine()
+            spinalVectorScore = (leftVectorScore + rightVectorScore) / 2.0
+        elif is_leftVector:
+            spinalVectorEstimate = pose_dix['left shoulder'], pose_dix['left hip']
+            # 10% score penalty in conficence
+            spinalVectorScore = leftVectorScore - (leftVectorScore * 0.1)
+        elif is_rightVector:
+            spinalVectorEstimate = pose_dix['right shoulder'], pose_dix['right hip']
+            # 10% score penalty in conficence
+            spinalVectorScore = rightVectorScore - (rightVectorScore * 0.1)
```
ivelin, line 270: @bhavikapanara maybe add a comment why we are penalizing the overall score. Also this can be slightly abbreviated as `rightVectorScore * 0.9` .

src/ambianic/pipeline/ai/fall_detect.py
```diff
@@ -261,17 +311,20 @@ def fall_detect(self, image=None):
...
                     log.info("The body-line angle with vertical axis is decreasing from the previous frame.")
                 else:
-                    angle_change = self.find_changes_in_angle(pose_dix)
+                    leaning_angle = self.find_changes_in_angle(pose_dix)
 
-                    if angle_change > self._fall_factor:
+                    leaning_probability = 1 if leaning_angle > self._fall_factor else 0
+                    fall_score = leaning_probability * (self.previous_body_vector_score + current_body_vector_score) / 2
+                    
+                    if leaning_angle > self._fall_factor:
```
ivelin, line 319: @bhavikapanara this is perfectly fine, but technically for clarity we should probably use
`if fall_score >= confidence_threshold` which does the same as the current check `leaning_angle > self._fall_factor` but is a bit easier to read.

ivelin (changes requested):
@bhavikapanara great work. This PR feel almost ready to go. Please see the few minor comments I made and let me know your thoughts.

src/ambianic/pipeline/ai/fall_detect.py
bhavikapanara: thanks @ivelin

src/ambianic/pipeline/ai/fall_detect.py
bhavikapanara: Yes, @ivelin  you are correct...it is redundant call.

src/ambianic/pipeline/ai/fall_detect.py
bhavikapanara: Done

src/ambianic/pipeline/ai/fall_detect.py
```diff
@@ -240,15 +294,12 @@ def fall_detect(self, image=None):
                 log.info("No pose with key-points found.")
                 return inference_result, thumbnail
             else:
-                pose_dix = {}
                 inference_result = []
+                pose_score, pose_dix = self.estimate_spinalVector_score(pose)
```
ivelin, line 298: @bhavikapanara why are we calling `estimate_spinalVector_score` here? The spinal score was already calculated a few lines above in `pose, thumbnail = self.find_keypoints(image)`. Shouldn't we just return the score in `find_keypoints` and reuse it?

ivelin (changes requested):
Good work. I only have one more comment, which I should have made earlier, but didn't spot the redundancy until now. 

Since we are on different time zones, please feel free to address the last comment and continue on to https://github.com/ambianic/ambianic-edge/issues/282 and https://github.com/ambianic/ambianic-edge/issues/283.

src/ambianic/pipeline/ai/fall_detect.py
bhavikapanara: yes..got it. Done

bhavikapanara:
> Good work. I only have one more comment, which I should have made earlier, but didn't spot the redundancy until now.
> 
> Since we are on different time zones, please feel free to address the last comment and continue on to #282 and #283.

I have already added attached screenshots of #283  in our test_fall_detect.py. And currently working on the #282 .

PR #296, opened 2021-02-04
fix: pose engine coordinates in inference output; closes #254

### LINKED ISSUE #254: [BUG] Pose Engine returns pose coordinates as original image pixel coordinates
**Describe the bug**
Currently pose engine returns pose keypoint coordinates as pixel coordinates in the original input image.
This is not ideal, because we send over the wire to the UI timeline just a thumbnail image, which is a scaled down lighter version of the original image.

To be consistent with other detection models, pose engine should return key points as relative values from 0 to 1 and let the visualization component calculate pixel coordinates right before they are shown to the user.

For privacy concerns, we may not even show the original image or thumbnail to some users, but instead only the stick figure with the pose keypoints.

**Additional context**
[See here](https://github.com/ambianic/ambianic-edge/blob/17873da6fc7d61ce0ff3180d0db4bbc0c0837c84/src/ambianic/pipeline/ai/pose_engine.py#L128).

bhavikapanara:


src/ambianic/pipeline/store.py
```diff
@@ -131,6 +131,71 @@ def _save_sample(self,
         self.notify(save_json)
         return image_path, json_path
 
+    def _save_sample_new(self,
```
bhavikapanara, line 134: @ivelin 

As I am not enabled to test it on UI, doesn't know the impact on UI, here I have added sample function for fall detection inference result processing. 

Please take a look at it. Is it the correct way?

ivelin:
I think we need to enable you to test with the UI. Let's schedule a call for early next week to walk you through the steps. How about Tuesday morning your time (Monday night for me in Austin)?

ivelin:
@bhavikapanara please note the format of PR titles makes a difference.

Instead of 

```
fix: fall detection issue #254
```

Please use this format that both triggers a link to the issue and is more descriptive for peer reviewers.

```
fix: closes #254; pose engine coordinates in inference output
```

ivelin:
> [PEP 8](https://www.python.org/dev/peps/pep-0008)

@bhavikapanara please address these PEP8 formatting issues because they are distracting during the review process.

You may want to look at [installing a PEP8 lint](https://code.visualstudio.com/docs/python/linting) in your dev environment so you see these early during original file edit.

ivelin (changes requested):
@bhavikapanara let me know how your testing goes with the UI.

src/ambianic/pipeline/store.py
ivelin: @bhavikapanara were you able that this change does not break the UI?

src/ambianic/pipeline/store.py
```diff
@@ -131,6 +131,71 @@ def _save_sample(self,
...
+        image_file = time_prefix.format(suffix='image', fext='jpg')
+        image_path = self._output_directory / image_file
+        thumbnail_file = time_prefix.format(suffix='thumbnail', fext='jpg')
+        thumbnail_path = self._output_directory / thumbnail_file
+        json_file = time_prefix.format(suffix='inference', fext='json')
+        json_path = self._output_directory / json_file
+        inf_json = []
+        if inference_result:
+            for inf in inference_result:
+                label, confidence, box, leaning_angle, keypoint_corr = inf
```
ivelin, line 150: @bhavikapanara as discussed on our call, the model specific inference classes (fall_detect.py, object_detect.py, etc.) should provide an inference object that is readily [json serializable](https://stackoverflow.com/questions/3768895/how-to-make-a-class-json-serializable)
.

store.py should just be left with the responsibility to **store** the inference data. It should just dump the provided inference data structure to json with one line of code:

```
inf_json.append(one_inf)
```

store.py should not have to know the details of formatting all possible types of inference outputs. As we add more inference types in the future, store.py will get too bloated and harder to maintain.

For now we can leave the logic in store.py to lookup a thumnail attribute and save it to disk along with the json dump, but in a next iteration I would like you to propose a more generic way to handle that. For example fall detection should save the pair of images that were used to determine a positive detection. We need them saved for debugging and also to present to users , so they are aware what the ground truth is (start of fall, end of fall).

ivelin (changes requested):

ivelin:
@bhavikapanara can we close this PR as an obsolete duplicate of #314 ?

PR #314, opened 2021-02-15
fix: separate inference output formatting for each detection model; closes #254

### LINKED ISSUE #254: [BUG] Pose Engine returns pose coordinates as original image pixel coordinates

bhavikapanara:


ivelin:
@bhavikapanara I did a first pass review and I think conceptually you captured the essence of the issue. I will do a second pass tomorrow and look for any details that I might have overlooked today before merging. Thank you!

ivelin (changes requested):
@bhavikapanara notice that the CI fails with testsuite errors. 

Please make sure all CI checks pass. 

Also please take care of the pep8 lint errors. I suggest installing pylama or another local linter in your IDE.

Thank you!

ivelin:
@bhavikapanara did you test these changes against the UI. All looking good in the timeline view?

bhavikapanara:
> @bhavikapanara did you test these changes against the UI. All looking good in the timeline view?

@ivelin I have checked against UI and it is working

bhavikapanara:
> @bhavikapanara notice that the CI fails with testsuite errors.
> 
> Please make sure all CI checks pass.
> 
> Also please take care of the pep8 lint errors. I suggest installing pylama or another local linter in your IDE.
> 
> Thank you!

@ivelin I have tried to fix all pep8 error...but still, there are some lines left. will do it.
And also working to resolve CI fails

bhavikapanara:
@ivelin CI test fails for test_avsource.py file.

Working to solve it.

Issue #330, opened 2021-02-18
[CRITICAL BUG] event serialization error in store.py


ivelin:
**Describe the bug**
After merging PR #314 , there is an exception in store.py when it tries to save a detection event.

```
ambianic-edge    | 2021-02-18 19:29:44 INFO /opt/ambianic-edge/src/ambianic/pipeline/ai/fall_detect.py.convert_inference_result(482): label: FALL , confidence: 1, leaning_angle: 79,                          keypoint_corr: {'left shoulder': [102.0, 86.0], 'left hip': [110.0, 95.0], 'right shoulder': [89.0, 94.0], 'right hip': [102.0, 97.0]}
ambianic-edge    | 2021-02-18 19:29:44 ERROR /opt/ambianic-edge/src/ambianic/pipeline/store.py.process_sample(159): Error TypeError('Object of type float32 is not JSON serializable') while saving sample {'image': <PIL.JpegImagePlugin.JpegImageFile image mode=RGB size=1280x720 at 0xADAA06F0>, 'thumbnail': <PIL.Image.Image image mode=RGB size=257x145 at 0xADA74E30>, 'inference_result': [{'label': 'FALL', 'confidence': 0.5335016838548498, 'leaning_angle': 79.3426890672073, 'keypoint_corr': {'left shoulder': [102.0, 86.0], 'left hip': [110.0, 95.0], 'right shoulder': [89.0, 94.0], 'right hip': [102.0, 97.0]}}], 'inference_meta': {'display': 'Fall Detection'}}
ambianic-edge    | Traceback (most recent call last):
ambianic-edge    |   File "/opt/ambianic-edge/src/ambianic/pipeline/store.py", line 143, in process_sample
ambianic-edge    |     inference_meta=inference_meta)
ambianic-edge    |   File "/opt/ambianic-edge/src/ambianic/pipeline/store.py", line 110, in _save_sample
--
ambianic-edge    |   File "/usr/lib/python3.7/json/encoder.py", line 179, in default
ambianic-edge    |     raise TypeError(f'Object of type {o.__class__.__name__} '
ambianic-edge    | TypeError: Object of type float32 is not JSON serializable
ambianic-edge    | 2021-02-18 19:29:44 INFO /opt/ambianic-edge/src/ambianic/pipeline/ai/tf_detect.py.log_stats(178): FallDetector inference time 0.02 ms, 6.72 fps in pipeline area_watch
[...]

PR #331, opened 2021-02-19
fix: event serialization error in store.py [issue #330]

bhavikapanara:
solved issue #330


ivelin (changes requested):
@bhavikapanara this is one line change to a code that already works. You need to resubmit the whole PR from #314 since it was all reverted to the code state before the PR was merged. As I asked, you need to cover thoroughly the new json serialization and store code with new tests that exercise the specific error we saw in production to be sure it does not happen again.

ivelin:
@bhavikapanara should this PR be closed as obsolete?

ivelin:
> @bhavikapanara should this PR be closed as obsolete?

@bhavikapanara did you see this question?

PR #332, opened 2021-02-19
fix: separate inference output formatting for each detection model; [issues #330, #254]

bhavikapanara:
Fix issues: #254 #330

### DIFF
--- a/src/ambianic/pipeline/ai/face_detect.py
+++ b/src/ambianic/pipeline/ai/face_detect.py
@@ -93,12 +93,12 @@ class FaceDetector(TFBoundingBoxDetection):
                          box)
                 one_inf = {
                     'label': label,
-                    'confidence': float(confidence),
+                    'confidence': confidence,
                     'box': {
-                        'xmin': float(box[0]),
-                        'ymin': float(box[1]),
-                        'xmax': float(box[2]),
-                        'ymax': float(box[3]),
+                        'xmin': box[0],
+                        'ymin': box[1],
+                        'xmax': box[2],
+                        'ymax': box[3],
                     }
                 }
                 inf_json.append(one_inf)
--- a/src/ambianic/pipeline/ai/face_detect.py
+++ b/src/ambianic/pipeline/ai/face_detect.py
@@ -16,10 +16,10 @@ class FaceDetector(TFBoundingBoxDetection):
         width, height = image.size
 
         # Setting the points for cropped image
-        left = box[0]*width
-        top = box[1]*height
-        right = box[2]*width
-        bottom = box[3]*height
+        left = box['xmin']*width
+        top = box['ymin']*height
+        right = box['xmax']*width
+        bottom = box['ymax']*height
 
         # Cropped image of above dimension
         # (It will not change orginal image)
@@ -46,9 +46,13 @@ class FaceDetector(TFBoundingBoxDetection):
                 else:
                     # - apply face detection to cropped person areas
                     # - pass face detections on to next pipe element
-                    for label, confidence, box in prev_inference_result:
+                    for e_result in prev_inference_result:
+                        label, confidence, box = e_result['label'], \
+                                                 e_result['confidence'], \
+                                                 e_result['box']
+
                         if label == 'person' and \
-                          confidence >= self._tfengine.confidence_threshold:
+                           confidence >= self._tfengine.confidence_threshold:
                             person_regions.append(box)
                     log.debug('Received %d person boxes for face detection',
                               len(person_regions))
@@ -56,6 +60,9 @@ class FaceDetector(TFBoundingBoxDetection):
                         person_image = self.crop_image(image, box)
                         thumbnail, tensor_image, inference_result = \
                             self.detect(image=person_image)
+
+                        inference_result = self.convert_inference_result(
+                            inference_result)
                         log.debug('Face detection inference_result: %r',
                                   inference_result)
                         inf_meta = {
@@ -73,3 +80,27 @@ class FaceDetector(TFBoundingBoxDetection):
                               'Dropping sample: %r',
                               e,
                               sample)
+
+    def convert_inference_result(self, inference_result):
+
+        inf_json = []
+        if inference_result:
+            for inf in inference_result:
+                label, confidence, box = inf[0:3]
+                log.info('label: %s , confidence: %.0f, box: %s',
+                         label,
+                         confidence,
+                         box)
+                one_inf = {
+                    'label': label,
+                    'confidence': float(confidence),
+                    'box': {
+                        'xmin': float(box[0]),
+                        'ymin': float(box[1]),
+                        'xmax': float(box[2]),
+                        'ymax': float(box[3]),
+                    }
+                }
+                inf_json.append(one_inf)
+
+        return inf_json
--- a/src/ambianic/pipeline/store.py
+++ b/src/ambianic/pipeline/store.py
@@ -185,9 +185,9 @@ class JsonEncoder(json.JSONEncoder):
     def default(self, obj):
         if isinstance(obj, np.integer):
             return int(obj)
-        elif isinstance(obj, np.floating):
+        if isinstance(obj, np.floating):
             return float(obj)
-        elif isinstance(obj, np.ndarray):
+        if isinstance(obj, np.ndarray):
             return obj.tolist()
-        else:
-            return super(JsonEncoder, self).default(obj)
+
+        return super(JsonEncoder, self).default(obj)
--- a/src/ambianic/pipeline/store.py
+++ b/src/ambianic/pipeline/store.py
@@ -5,7 +5,7 @@ import pathlib
 import json
 import uuid
 from typing import Iterable
-
+import numpy as np
 from ambianic import DEFAULT_DATA_DIR
 from ambianic.pipeline import PipeElement
 from ambianic.notification import Notification, NotificationHandler
@@ -105,7 +105,8 @@ class SaveDetectionSamples(PipeElement):
         thumbnail.save(thumbnail_path)
         # save samples to local disk
         with open(json_path, 'w', encoding='utf-8') as f:
-            json.dump(save_json, f, ensure_ascii=False, indent=4)
+            json.dump(save_json, f, ensure_ascii=False, indent=4,
+                      cls=JsonEncoder)
         # e = PipelineEvent('Detected Objects', type='ObjectDetection')
         self.event_log.info('Detection Event', save_json)
         log.debug("Saved sample (detection event): %r ", save_json)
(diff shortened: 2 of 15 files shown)


ivelin (changes requested):
@bhavikapanara I don't see new test cases that specifically try to exercise and cover the code that caused the error.

ivelin:
@bhavikapanara look closer at the error again and see which method should have been caught by the test suite before, but it wasn't. That's the one you should write new tests for to ensure it does not fail from now on. 

Hint: `_save_sample`

Make sure that the tests exercise the code and check whether it produces the desired outcome on disk.

src/ambianic/pipeline/ai/face_detect.py
```diff
@@ -73,3 +80,27 @@ def process_sample(self, **sample):
...
+        if inference_result:
+            for inf in inference_result:
+                label, confidence, box = inf[0:3]
+                log.info('label: %s , confidence: %.0f, box: %s',
+                         label,
+                         confidence,
+                         box)
+                one_inf = {
+                    'label': label,
+                    'confidence': float(confidence),
```
ivelin, line 96: @bhavikapanara instead of looking for all the possible places that a numpy type serialization can occur and hardcoding a type conversion, which is error prone and not too elegant, I suggest looking at more robust solutions. For example finding a library that can handle json serialization of numpy types such as [this one](https://github.com/mverleg/pyjson_tricks). There may be other alternatives. Please research.

ivelin: Looks like [encoding numpy scalars is an issue](https://json-tricks.readthedocs.io/en/latest/#numpy-scalars) for json-tricks.

The numpy team is also [aware of the issue](https://github.com/numpy/numpy/issues/16432) and there is no agreed upon solution yet.

Looks like adding a simple custom encoder extension is the preferred solution at the moment as presented [1)here](https://stackoverflow.com/a/27050186/12015435) and  [2)here](https://gist.github.com/massgh/297a73f2dba017ffd28dbc34b9a40e90).

bhavikapanara: Ok...Will implement this

bhavikapanara: Here I have added the encoder to parse the numpy scalar.
https://github.com/ambianic/ambianic-edge/blob/4b0c350ab99df9089ae9a11a76fb6735d489267b/src/ambianic/pipeline/store.py#L108

For the test case, 
I have observed that the `_save_sample` function is covered by the try block, so whenever `json.dump()` fails, exception raise. So Assertion won't work here in test cases

https://github.com/ambianic/ambianic-edge/blob/4b0c350ab99df9089ae9a11a76fb6735d489267b/src/ambianic/pipeline/store.py#L157

ivelin: Encoder looks good.

For testing this type of cases, see the PipelineElement [mockup with callback registration](https://github.com/ambianic/ambianic-edge/blob/4b0c350ab99df9089ae9a11a76fb6735d489267b/tests/pipeline/ai/test_face_detect.py#L63) technique we use in the same test file. Think how you can implement a similar callback around `_save_sample` to make sure that the new encoder works in context.  

I would also suggest adding a few unit tests just around the encoder. A code that failed before with the default encoder (assert failure) should not fail with the new encoder (assert expected result).

ivelin:
@bhavikapanara There is 7.3% regression in code coverage in `store.py`. We should be increasing code coverage with each PR to prevent production bombs like the one this week.

```
src/ambianic/pipeline/store.py | 92.70% <41.66%> (-7.30%)
```



> # [Codecov](https://codecov.io/gh/ambianic/ambianic-edge/pull/332?src=pr&el=h1) Report
> > Merging [#332](https://codecov.io/gh/ambianic/ambianic-edge/pull/332?src=pr&el=desc) ([4b0c350](https://github.com/ambianic/ambianic-edge/commit/4b0c350ab99df9089ae9a11a76fb6735d489267b)) into [master](https://codecov.io/gh/ambianic/ambianic-edge/commit/ee79c40e019b82ed4dd1c89cf9cd2be40e51643f?el=desc) ([ee79c40](https://github.com/ambianic/ambianic-edge/commit/ee79c40e019b82ed4dd1c89cf9cd2be40e51643f)) will **decrease** coverage by `0.23%`.
> > The diff coverage is `91.89%`.
> 
> [Impacted file tree graph](https://codecov.io/gh/ambianic/ambianic-edge/pull/332?src=pr&el=tree)
> 
> ```diff
> @@            Coverage Diff             @@
> ##           master     #332      +/-   ##
> ==========================================
> - Coverage   94.53%   94.30%   -0.24%     
> ==========================================
>   Files          23       23              
>   Lines        2197     2230      +33     
> ==========================================
> + Hits         2077     2103      +26     
> - Misses        120      127       +7     
> ```
> 
> [Impacted Files](https://codecov.io/gh/ambianic/ambianic-edge/pull/332?src=pr&el=tree)	Coverage Δ	
> [src/ambianic/pipeline/store.py](https://codecov.io/gh/ambianic/ambianic-edge/pull/332/diff?src=pr&el=tree#diff-c3JjL2FtYmlhbmljL3BpcGVsaW5lL3N0b3JlLnB5)	`92.70% <41.66%> (-7.30%)`	⬇️
> [src/ambianic/pipeline/ai/fall_detect.py](https://codecov.io/gh/ambianic/ambianic-edge/pull/332/diff?src=pr&el=tree#diff-c3JjL2FtYmlhbmljL3BpcGVsaW5lL2FpL2ZhbGxfZGV0ZWN0LnB5)	`97.27% <96.22%> (+0.11%)`	⬆️
> [src/ambianic/pipeline/ai/face_detect.py](https://codecov.io/gh/ambianic/ambianic-edge/pull/332/diff?src=pr&el=tree#diff-c3JjL2FtYmlhbmljL3BpcGVsaW5lL2FpL2ZhY2VfZGV0ZWN0LnB5)	`100.00% <100.00%> (ø)`	
> [...mbianic/pipeline/ai/image_boundingBox_detection.py](https://codecov.io/gh/ambianic/ambianic-edge/pull/332/diff?src=pr&el=tree#diff-c3JjL2FtYmlhbmljL3BpcGVsaW5lL2FpL2ltYWdlX2JvdW5kaW5nQm94X2RldGVjdGlvbi5weQ==)	`100.00% <100.00%> (ø)`	
> [src/ambianic/pipeline/ai/object_detect.py](https://codecov.io/gh/ambianic/ambianic-edge/pull/332/diff?src=pr&el=tree#diff-c3JjL2FtYmlhbmljL3BpcGVsaW5lL2FpL29iamVjdF9kZXRlY3QucHk=)	`100.00% <100.00%> (ø)`	
> [src/ambianic/pipeline/ai/pose_engine.py](https://codecov.io/gh/ambianic/ambianic-edge/pull/332/diff?src=pr&el=tree#diff-c3JjL2FtYmlhbmljL3BpcGVsaW5lL2FpL3Bvc2VfZW5naW5lLnB5)	`97.93% <100.00%> (ø)`

src/ambianic/pipeline/ai/face_detect.py
bhavikapanara: Ok.
Working on this

src/ambianic/pipeline/store.py
```diff
@@ -197,3 +179,15 @@ def notify(self, save_json: dict):
             }
             notification = Notification(data=data, providers=self.notification_config["providers"])
             self.notification.send(notification)
+
+
+class JsonEncoder(json.JSONEncoder):
+    def default(self, obj):
+        if isinstance(obj, np.integer):
+            return int(obj)
```
bhavikapanara, line 187: @ivelin Here I got codefactor error. I don't know how to manage to fix it.

bhavikapanara: I have solved it

tests/pipeline/test_store.py
```diff
@@ -18,6 +58,87 @@ def test_process_sample_none():
...
+    print(json_dir)
+    print(store._json_path)
+    with open(store._json_path) as f:
+        json_inf = json.load(f)
+        print(json_inf)
+        img_fname = json_inf['image_file_name']
+        rel_dir = json_inf['rel_dir']
+        img_fpath = os.path.join(out_dir, rel_dir, img_fname)
+        assert img_fpath == str(store._img_path)
+        json_inf_res = json_inf['inference_result']
```
ivelin, line 131: @bhavikapanara let's also check that the image actually exists on disk.

tests/pipeline/test_store.py
```diff
@@ -18,6 +58,87 @@ def test_process_sample_none():
...
+        assert len(json_inf_res) == 1
+        json_inf_res = json_inf_res[0]
+        assert json_inf_res['label'] == 'person'
+        assert isinstance(json_inf_res['confidence'], float)
+        assert isinstance(json_inf_res['box']['xmin'], float)
+        assert isinstance(json_inf_res['box']['ymin'], float)
+        assert isinstance(json_inf_res['box']['xmax'], float)
+        assert isinstance(json_inf_res['box']['ymax'], float)
+
```
ivelin, line 140: @bhavikapanara let's clean up after the test.  Otherwise it may lead to side effects that affect other tests, github tracking, etc.

See [pytest fixtures](https://docs.pytest.org/en/latest/fixture.html#fixture-finalization-executing-teardown-code). 
Among other boilerplate fixtures, there is a convenient way to use per test tmp directory.

ivelin (changes requested):
@bhavikapanara Nice update. Please see a couple of fine tuning comments that should close this PR.

tests/pipeline/test_store.py
bhavikapanara: @ivelin Did you mean to delete the created directory (i.e tmp/detections/20210223-140610.026448) after finishing the test?

ivelin:
Yes

tests/pipeline/test_store.py
bhavikapanara: Done

tests/pipeline/test_store.py
bhavikapanara: Done

tests/pipeline/test_store.py
ivelin: OK, this works, but please take a note to move towards pytest fixtures going forward.

ivelin: @bhavikapanara  See how much cleaner the code would look with the pytest fixture for temp dirs:

https://docs.pytest.org/en/latest/reference.html#tmpdir

```
tmpdir
Tutorial: Temporary directories and files

tmpdir()[source]
Return a temporary directory path object which is unique to each test function invocation, created as a sub directory of the base temporary directory.

By default, a new base temporary directory is created each test session, and old bases are removed after 3 sessions, to aid in debugging. If --basetemp is used then it is cleared each session. See The default base temporary directory.

The returned object is a py.path.local path object.


```

ivelin (approved):

src/ambianic/pipeline/ai/face_detect.py
ivelin: OK, I am merging this PR. Please submit encoder unit tests as a new PR.

PR #391, opened 2021-11-02
feat: added default priority for all events to config file

vickywane:
This pull request will fix this [issue](https://github.com/ambianic/ambianic-ui/issues/687).

## Purpose
The `WARNING` priority adds a warning level to all detected events allowing them to be captured within the color formatter in the Ambianic PWA timeline.

## Approach
The new code changes within this pull request adds a default priority of `WARNING` level to the `config.yaml` file.

## Merge Checklist
- [x]  The code change is tested and works locally.
- [x]  The user and dev documentation is up to date.
- [x]  There is no commented out code in this PR.
- [x]  No lint errors (use flake8)
- [x]  100% test coverage

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -88,6 +88,7 @@ pipelines:
      - detect_falls: # look for falls
         ai_model: fall_detection
         confidence_threshold: 0.6
+        priority: WARNING # default priority for all event detections.
      - save_detections: # save samples from the inference results
         positive_interval: 10
         idle_interval: 600000
--- a/src/ambianic/pipeline/timeline_event.py
+++ b/src/ambianic/pipeline/timeline_event.py
@@ -5,6 +5,7 @@ import os
 import pathlib
 import uuid
 
+import ambianic
 import yaml
 from concurrent_log_handler import ConcurrentRotatingFileHandler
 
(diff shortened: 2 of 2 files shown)


ivelin (commented):
How does this priority propagate to the events log and then to the client? I don't see a linked test case or UI PR that demonstrates the effects of this change.

vickywane:
> How does this priority propagate to the events log and then to the client? I don't see a linked test case or UI PR that demonstrates the effects of this change.

I had to do some read through of how the `priority` field is being sent to the PWA, and I discovered that it comes from the `PipelineEventFormatter` class in the `timeline_event.py` file. 

Based on my understanding of the opened [issue](https://github.com/ambianic/ambianic-ui/issues/687), we need a way to store and retrieve a user's custom priority. Initially the record level from the `logging` package at a default of `INFO` was being used. 

Within this PR so far, I have readjusted the `priority` to be retrieved from the `config.yaml` file.

vickywane:
@ivelin Would it be okay to begin storing the priority level for event detection in a separate YAML file?

Can we use the structure below to store the priorities? 
```yaml
detections:
    fall: WARNING
    person: INFO
    cat: ERROR
```

ivelin:
Based on the roundtable call a couple of weeks ago, we need to reconsider what the end users need in terms of event prioritization, filtering and display.

